PR #32, opened 2016-04-04
Clean filenames before uploading

chrfritsch:
After having some issues with dirty image filenames, we want to clean filenames before uploading. @Cyberschorsch, @IT-Cru and @dbosen please review if this would fit for your needs.


dbosen:
Please add an admin setting to enable / disable this functionality or extract it to another module which can be enabled / disabled

PR #46, opened 2016-06-30
DKRFTN-695: fix php fatal while not having an node object in ThunderArticleBreadcrumbBuilder applies method

steffenr:
While using the **Revert Revision** function of the drupal core revisioning system, we get an fatal error coming from the ThunderArticleBreadcrumbBuilder applies method. In this case we don't have a node object. The PR fixes this behaviour by explicitly checking for an object.


chrfritsch:
Thx

PR #67, opened 2016-08-03
Integrate the tour module

chrfritsch:


dbosen:
Step 7 / 7 of article create tour says 
"The paragraphs section represents the main article content. Here you can your text and various types of media."
The text should reflect, that you cann see how paragraphs wor in the article edit tour and should also link to that tour

dbosen:
Step 5 / 7 of article create tour says:

"Teaser media
Enter the information that will be published at the front page."

This points to the Teaser media selection. So add something like:

"Teaser media
Enter the information that will be published at the front page. First upload or select a teaser image."

dbosen:
"Edit an existing article" Tour:
Do not repeat the new article tour. Maybe we should just rename it to "Introduction to paragraphs" and add examples for all available paragraphs.

PR #79, opened 2016-08-19
Streamline media teaser and entity browser, remove media_entity_browser

chrfritsch:


dbosen:
In Media Entity Browser rename "Create video" to "Embed Video", add description how to use the URL

PR #87, opened 2016-09-06
Feature/thunder updater

dbosen:
Pull request created, to be able to add code comments

### DIFF
--- a/modules/thunder_updater/src/Updater.php
+++ b/modules/thunder_updater/src/Updater.php
@@ -3,31 +3,27 @@
 namespace Drupal\thunder_updater;
 
 use Drupal\config_update\ConfigListInterface;
-use Drupal\config_update\ConfigReverter;
 use Drupal\config_update\ConfigRevertInterface;
-use Drupal\Core\Config\ConfigFactoryInterface;
-use Drupal\Core\Config\FileStorage;
-use Drupal\Core\Config\NullStorage;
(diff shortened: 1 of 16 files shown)


modules/thunder_updater/src/Updater.php
```diff
@@ -0,0 +1,484 @@
...
+          );
+        }
+
+        $updatePatch[$configName->getFullName()] = $updateDiff->getEdits();
+      }
+    }
+
+    if (!empty($updatePatch)) {
+      $patchFilePath = $this->getPatchFileName($module_name, $version_name);
+      file_put_contents($patchFilePath, gzencode(serialize($updatePatch)));
```
dbosen: This fails, if the update directory has not been created before, can druapls file_save_data be used for this instead?

dbosen:
I did the following:

- install thunder
- enable thunder_updater, but change nothing else
- drush cex -> this creates the initial set of configuration files
- call thunder-updater/generate-patch/thunder_article/beta12 -> this creates a unexpected large diff, propably due to reordering of the config keys?
- call thunder-updater/apply-patch/thunder_article/beta12 -> this updates lots of configuration
- call drush cex -> equally large export of changes
- call drush cim -> confirms that exported configuration is equal to active config
- call thunder-updater/generate-patch/thunder_article/beta13 -> this again creates a very large patch file, why?

PR #95, opened 2016-09-09
Taxonomy term status had no effect, added an access handler

chrfritsch:


dbosen:
Do we have to warn our brands, that this might change the shown terms?

chrfritsch:
We should do this on the next DDD and we should add a warning in the release notes

PR #98, opened 2016-09-12
Move paragraphs config into separate modle

chrfritsch:
And use paragraphs on term pages


dbosen:
Investigate, if it is possible to create an update hook in thunder_article, to make the transition easier

PR #108, opened 2016-09-20
#2802845 Move template files from thunder_article to appropriate modules

steffenr:
* https://www.drupal.org/node/2802845


goldquest:
This won't work, since we had to add the templates directory in the thunder_article manually to the theme engine. The problem was, that templates in modules won't override other templates. Only themes do that. 
See: https://github.com/BurdaMagazinOrg/thunder-distribution/blob/8.x-1.x/modules/thunder_article/thunder_article.module#L45

goldquest:
I don't even know, if we should move these to the corresponding modules, since we will have modules, we don't have access to.
I would agree on a thunder_fia module, which handles all the fia stuff for thunder.

steffenr:
Ah - ok - i was looking through my "themer" glasses ;) - sorry. 
As you suggested we should move these files to a more specific module, handling all the fbia stuff. In this case we could seperate things in a clearer way.

steffenr:
@goldquest I'll close the PR and will add your comments to the d.o. issue. Would be great to split up the non-directly related templates / functionality into a separate module.

chrfritsch:
Here is a PR for a thunder_fia module https://github.com/BurdaMagazinOrg/thunder-distribution/pull/111

PR #109, opened 2016-09-21
Add metatags for article

chrfritsch:


dbosen:
Update hooks are missing. thunder media update should install the image styles, thunder_article and the distribution itself should update the corresponding configurations

PR #111, opened 2016-10-07
Move alle the fb instant article stuff to a seperate module

chrfritsch:
@goldquest and @dbosen please check if this might work. I've no clue how to test the instant articles stuff


dbosen:
update hook is missing, that activates the new module if necessary

PR #112, opened 2016-10-07
Introduce media expire

chrfritsch:


dbosen:
Please provide a README.md explaining how to use this

dbosen:
Default image is not shown when original image is expired. a) When user is not logged in, nothing is shown. b) when user is logged in and the image is part of a gallery, the expired image is shown instead of the default image

PR #117, opened 2016-10-11
updated infinite-theme css

goldquest:
The sass files in infinite-theme weren't compiled for some months. Compiled them and fixed social bars in the header.


dbosen:
The composer file is still using the old tagged release of infinite theme. It would be best, if you tagged the infinite theme first and then use this tag in the drush make files as well.

PR #134, opened 2016-10-24
Add author filter to content overview

chrfritsch:


dbosen:
Update hook is missing that adds the exposed filter to existing installations

PR #144, opened 2016-11-07
remove theme from profile, add to composer and make files

pixelmord:
This needs review and testing. I use composer and local paths which works fine for development, however I could not get it to work with the remote repo for thunder_admin, my first guess was that the repo name and name in composer.json causes that problem, even though I explicitly added the git repo to the repositories array.
Makefiles I did not test.


pixelmord:
This PR was based off of hotfix/next branch and that is not the desired workflow

PR #150, opened 2016-11-11
Make instagrams responsive

chrfritsch:


mtodor:
We need styles adjustment. With fixed show/hide selection functionality, buttons are not properly aligned when selection is hidden.

PR #153, opened 2016-11-14
Check for existence of needed array keys to prevent php notice

steffenr:
In some cases we get an PHP notice while creating new media entities. This PR checks for existence of $pathinfo['filename'] and $pathinfo['extension'] in thunder_media module.


dbosen:
Would it be better to check for empty() instead of isset()?

steffenr:
@dbosen I changed the PR according to your last comment. By using !empty() it should be "more secure".

dbosen:
Ok, thanks. Now make sure, that 
phpcs --standard=Drupal --extensions=php,module,inc,install,test,profile,theme -p .
and
phpcs --standard=DrupalPractice --extensions=php,module,inc,install,test,profile,theme -p .
do not fail. 
More information on using that: https://github.com/BurdaMagazinOrg/thunder-dev-tools/blob/master/README.md#code-style-guidelines

dbosen:
I could push to your repository, is this correct? ;) Anyway, i fixed the coding style issue

steffenr:
Thanks.

dbosen:
I merged it into hotfix/next we will create a new rc from that soon

PR #162, opened 2016-11-21
Render author using a view mode specific template

freudenberg:


modules/thunder_article/modules/thunder_fia/templates/user--facebook-instant-articles-rss.html.twig
```diff
@@ -0,0 +1 @@
+<a>{{ content.field_forename|render|trim }} {{ content.field_surname|render|trim }}</a>
```
chrfritsch: There are no field_forename and field_surname fields in Thunder. Please use just the username

chrfritsch (commented):
We need an update hook to install the view mode on existing sites. An example can be found in thunder_update_8006

freudenberg:
Using a user view mode for rendering the FIA author tag makes the module independent of the fields attached to user entities. If thunder had a set of default fields it would make sense to provide a configuration for the FIA view mode.

freudenberg:
Why has this been closed? It's important.

dbosen:
This was an automatic close from github, when the base branch was merged, please reopen it against the 8.x-1.x branch.
As @chrfritsch already stated, we still need an update hook for this.

modules/thunder_article/modules/thunder_fia/templates/user--facebook-instant-articles-rss.html.twig
```diff
@@ -0,0 +1 @@
+<a>{{ user.displayName }}</a>
```
goldquest: If we use a new viewmode for the user, shouldn't we then also render the full user?

goldquest: The problem is, that in the default configuration, the User entity doesn't have any name associated. Only a picture and "Member for" would be displayed.

goldquest:
Another option would be to keep the `user.displayName` and implement a hook to modify it: https://api.drupal.org/api/drupal/core!modules!user!user.api.php/function/hook_user_format_name_alter/8.2.x

dbosen (changes requested):
We need an update hook which imports the new configuration into the active configuration.
This should be done with 'config_update.config_update' service.
For examples on how to do this take a look here: https://github.com/BurdaMagazinOrg/thunder-distribution/blob/a84286911131c685173222378c79b7b59eb2e138/modules/thunder_media/thunder_media.install in function thunder_media_update_8102()

freudenberg:
Please check the last commit.

chrfritsch:
There are Coding Style issues. You can see them in the travis log https://travis-ci.org/BurdaMagazinOrg/thunder-distribution/jobs/199575911

PR #164, opened 2016-11-22
Feature/blazy responsive

goldquest:


modules/thunder_media/config/install/responsive_image.styles.media_image.yml
```diff
@@ -0,0 +1,29 @@
+langcode: en
+status: true
+dependencies:
+  config:
+    - image.style.media_image
+    - image.style.media_image_mobile
+    - image.style.media_image_tablet
+#  theme:
+#    - bartik
```
goldquest: With this dependency enabled, the installation breaks. But since bartik is installed automatically, I think it is not an issue.

chrfritsch: I think the responsive image style should not depend on a theme. Couldn't we use "Breakpoint group" "Responsive image" for that?

PR #166, opened 2016-11-22
Adjusted few update hooks that could fail

mtodor:
Adjusted few update hooks that failed on update testing.
Also adjusted configuration files with "_core.default_config_hash".


chrfritsch:
The config_update is not enabled in thunder. You should enable it in your update hook.

And then add it to thunder.info.yml that we didn't need to enable it in later update hooks.

PR #170, opened 2016-11-23
Fixed checking for required view mode in update hook

mtodor:


mtodor:
Added change to test "update" testing, it should be removed before pull request is accepted.

mtodor:
Pull request is ready now.

PR #173, opened 2016-11-25
fix update

dbosen:


mtodor:
Adjusted message to be something like this: "Unable to import entity_view_display.media.instagram.thumbnail config."

mtodor:
Pull request is ready for merge.

PR #176, opened 2016-11-27
#2830803 Avoid using fixed versions in composer.json

fago:
See https://www.drupal.org/node/2830803.

Requiring fixed versions in composer.json should be avoided. Always use the ^ parameter with it instead to allow people to use future versions or dev versions as necessary.

The PR fixes the problem and allows me to do
composer require drupal/paragraphs=1.x-dev
in my project.


chrfritsch:
We have to use fixed versions for projects where we ship some patches

dbosen:
Works as intended, we cannot remove th fixed versions on modules that need patches

PR #179, opened 2016-11-28
Update hooks

mtodor:
Additional fixing for update hooks:

- missing "use"
- missing output of log
- and catching all exceptions


mtodor:
-> PR is not ready, because there are changes to trigger update test.

mtodor:
PR is ready now

PR #189, opened 2016-12-06
Remove useless filter, add filter to search in galleries

chrfritsch:


chrfritsch:
I could not provide an update hook because the change in the view is to complex

dbosen:
Removing the language and changing the default publlished filter is good. But the gallery filter without a good autocomplete for entity references is not good. 
For now just remove the gallery field from this PR and keep the rest. 
Please write an update hook that removes the language filter from existing installations.

chrfritsch:
This also needs a checkbox for thunder updater and update hook should start with thunder_media_81xx

PR #195, opened 2016-12-13
use paragraph dev version while developing

dbosen:
I switched to paragraphs dev branch for developing purposes, since the last released version is quite dated. This should be changed back to the latest released version when merging this into the main development branch of the distribution.


dbosen:
@pixelmord do you agree on this? If so, please review and merge

pixelmord:
I agree, I did require the dev version already for the feature branch I was working on, but it makes sense to have this in this base/demo branch already

PR #197, opened 2016-12-20
Integrate social buttons

chrfritsch:


dbosen:
The twitter link opens two popups on firefox, this has to be investigated before merging this.

chrfritsch:
This happens in firefox for articles which have a twitter widget in their content

chrfritsch:
I think this is an JS issue

goldquest:
The problem is, that the twitter widget library still gets loaded. It reacts on clicks on links, which have an intent URL. See example here: http://jsbin.com/yaniguriri/1/edit?html,output
We have to remove the twitter library (which is actually, why shariff exists), or find a way to disable the twitter library for these links.

goldquest:
AFAICS there is no possibility to have some attribute in the link, to disable it.

goldquest:
Ok, I think the best solution is to stop the event bubbling on shariff links

goldquest:
The code would look something like this:
```
jQuery('.shariff').click(function(e) {
  e.stopPropagation()
})
```

chrfritsch:
If this fixes our problem, i am fine with that

chrfritsch:
This also needs a checkbox in thunder updater

Issue #199, opened 2016-12-30
Demo content bugs


landsman:
Hi,

1) This is screenshot is from today fresh install (infinite theme, demo content):
![fresh-install](https://cloud.githubusercontent.com/assets/3184228/21567796/533a4684-ceaf-11e6-89a7-9dccc2fb2167.png)
... after reload exception is gone but this is looking very oddly.

2) In demo content is bug about ajax paginator. 
Requested url is "/home" but this is url or alias not exist. Home is in code(!) here: https://github.com/BurdaMagazinOrg/infinite-module/blob/8.x-1.x/src/Plugin/Block/InfiniteLazyLoadingSimple.php#L54

![snimek obrazovky 2016-12-30 v 16 47 32](https://cloud.githubusercontent.com/assets/3184228/21567845/c98aa5ea-ceaf-11e6-8efe-cbe719d48186.png)

Is possible create more stable demo content for new members?

**Environment:**
- OS X 10.12.2 (16C67)
- MAMP PRO with OPcache
- npm 3.10.8
- php 7.0.12
- mysql 5.6.33

goldquest:
Thanks for the issue. Will have a look at it.

goldquest:
Blazy beta6 is not compatible with slick at the moment. See Drupal issue here: https://www.drupal.org/node/2841316.

landsman:
OK, next thing is demo data setup.

landsman:
I don't know how this exception can related with Blazy&Slick issue but this shows after few reloads on article detail - demo content:

`Notice: Array to string conversion in Drupal\infinite_base\Plugin\Block\InfiniteLazyLoadingSimple->build() (line 68 of modules/contrib/infinite_module/src/Plugin/Block/InfiniteLazyLoadingSimple.php).
Drupal\infinite_base\Plugin\Block\InfiniteLazyLoadingSimple->build() (Line: 203)
Drupal\block\BlockViewBuilder::preRender(Array)
call_user_func('Drupal\block\BlockViewBuilder::preRender', Array) (Line: 376)
Drupal\Core\Render\Renderer->doRender(Array, ) (Line: 195)
Drupal\Core\Render\Renderer->render(Array) (Line: 474)
Drupal\Core\Template\TwigExtension->escapeFilter(Object, Array, 'html', NULL, 1) (Line: 58)
[...]

PR #200, opened 2017-01-05
Blazy beta6 is not compatible with slick at the moment

goldquest:
Because of an incompatibility between blazy beta6 and slick, we have to lock the blazy version to beta5. 
Drupal Issue: https://www.drupal.org/node/2841316
Our GitHub Issue: #199 

Maybe we should think again about how we solve the issue, that we don't want to lock people to module versions, but still provide a stable distro version.
Maybe locking the versions for "stable" releases and letting them open for "dev"?


landsman:
And why are you using Slick?
Is not better use the Owl Carousel for example?

PR #207, opened 2017-01-13
Introduce thunder_updater to support thunder users by doing updates

chrfritsch:


thunder.install
```diff
@@ -305,3 +305,15 @@ function thunder_update_8102() {
...
+/**
+ * Installs the thunder_updater module.
+ */
+function thunder_update_8103() {
+
+  \Drupal::service('module_installer')->install([
+    'thunder_updater',
+  ]);
+
+  \Drupal::service('thunder_updater')->checkAllListPoints(FALSE);
```
mtodor: We need bullet-points for hooks: 8101 and 8102. Then this hook should be executed before 8101 and 8102, because bullet-points that should be checked by listed hooks should not be cleared by this hook. We can add "hook_update_dependencies" (and list that 8101 and 8102 depend on 8103). Other possible solution is to exchange hooks 8101 and 8103, so that order is correct.

thunder.install
```diff
@@ -311,9 +312,51 @@ function thunder_update_8102() {
...
+
+  $moduleInstaller = \Drupal::service('module_installer');
+
+  try {
+    if (!$moduleInstaller->install(['thunder_updater'])) {
+      $updateLogger->warning(t('Unable to enable the Thunder Updater'));
+    }
+  }
+  catch (MissingDependencyException $e) {
+    $updateLogger->warning(t('Unable to enable the Thunder Updater because of missing dependencies.'));
```
mtodor: If thunder updater module is not installed, then we can't use service provided by it, that means that rest of execution is not possible -> so we should simply just return from update hook.

thunder.install
```diff
@@ -311,9 +312,51 @@ function thunder_update_8102() {
...
+  }
 
   \Drupal::service('thunder_updater')->checkAllListPoints(FALSE);
```
mtodor: Thunder updater service should be assigned to variable -> and then used on multiple places.

thunder.install
```diff
@@ -311,9 +312,51 @@ function thunder_update_8102() {
...
+
+  $updates = [
```
mtodor: Maybe we should document this part, that it's related to other two hooks which are executed before, just to mark existing checkpoints properly.

PR #210, opened 2017-01-17
Update infinite

chrfritsch:


chrfritsch:
Currently no blazy/slick update. It fails, we have to investigate more

PR #212, opened 2017-01-19
Update entity browser

chrfritsch:


chrfritsch:
We should keep our configuration clean. So i added an update hook to remove the auto_select key where its not supported

modules/thunder_media/thunder_media.install
```diff
@@ -338,3 +339,21 @@ function thunder_media_update_8010() {
 
   return $updateLogger->outputHtml();
 }
+
+/**
+ * Remove auto_select from EB widget config, for those who dont't support it.
```
mtodor: There is typo here "dont't" -> I'm also not sure can we formulate update message to be more understandable for user: fe. Cleanup of "auto_select" option for entity browser widget configurations.
Or something similar, so that it's more clear what update hook is doing. I'm not sure what is best explanation.

PR #214, opened 2017-01-19
Add a save and continue button to the node form

chrfritsch:

### DIFF
--- a/modules/thunder_article/src/ThunderNodeForm.php
+++ b/modules/thunder_article/src/ThunderNodeForm.php
@@ -19,7 +19,7 @@ class ThunderNodeForm extends NodeForm {
     if ($element['publish']['#access'] && \Drupal::currentUser()->hasPermission('administer nodes')) {
       $element['save_continue'] = $element['publish'];
       $element['save_continue']['#value'] = t('Save and continue');
-      $element['save_continue']['#weight'] = $element['publish']['#weight'] - 1;
+      $element['save_continue']['#weight'] = min($element['publish']['#weight'], $element['unpublish']['#weight']) - 1;
     }
 
     return $element;
--- a/modules/thunder_article/src/ThunderNodeForm.php
+++ b/modules/thunder_article/src/ThunderNodeForm.php
@@ -1,0 +1,40 @@
+<?php
+
+namespace Drupal\thunder_article;
+
+use Drupal\Core\Form\FormStateInterface;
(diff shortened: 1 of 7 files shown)


tests/src/FunctionalJavascript/ThunderJavascriptTestBase.php
```diff
@@ -284,4 +284,23 @@ public function runCron() {
       ->click();
   }
 
+  /**
+   * Click article save option based on index of action.
```
mtodor: I think we should document options here -> it will help us in future when we use this method.

Some list:
1 - Save and continue
2 - Save as unpublished (default)
3 - Save and publish

modules/thunder_article/src/ThunderNodeForm.php
```diff
@@ -0,0 +1,40 @@
...
+class ThunderNodeForm extends NodeForm {
+
+  /**
+   * {@inheritdoc}
+   */
+  protected function actions(array $form, FormStateInterface $form_state) {
+    $element = parent::actions($form, $form_state);
+
+    if ($element['publish']['#access'] && \Drupal::currentUser()->hasPermission('administer nodes')) {
+      $element['save_continue'] = $element['publish'];
```
mtodor: When you create new article and you click "Save and continue" -> Article will be automatically published. I'm not sure is that wanted behavior, I think article should stay unpublished by default. Also I'm not sure how good is this option during editing of already published article, because changes will be published.

Anyway, when article is edited, "Save and continue" should not change state of article (published, unpublished). When you click "Save and continue" on unpublished article it should stay unpublished, if you click same option on published article it should stay published.

modules/thunder_article/src/ThunderNodeForm.php
```diff
@@ -0,0 +1,40 @@
...
+ */
+class ThunderNodeForm extends NodeForm {
+
+  /**
+   * {@inheritdoc}
+   */
+  protected function actions(array $form, FormStateInterface $form_state) {
+    $element = parent::actions($form, $form_state);
+
+    if ($element['publish']['#access'] && \Drupal::currentUser()->hasPermission('administer nodes')) {
```
mtodor: This should be more strict checking if "$element['publish']" doesn't exist. Also for:
`min($element['publish']['#weight'], $element['unpublish']['#weight'])`.

PR #225, opened 2017-01-31
Feature/amp

goldquest:
This adds a new subtheme to use AMP with Thunder

To test it, you have to activate the AMP module, install the `thunder_amp` theme and go to an article. Here you have to append `?amp` to the URL.
To validate the output you have different options. Please read about them here: https://www.ampproject.org/docs/guides/validate

- [ ] ~Necessary update hooks are provided.~
- [ ] ~User roles have correct access for new introduced permission.~


config/optional/amp.theme.yml
```diff
@@ -0,0 +1 @@
+amptheme: thunder_amp
```
goldquest: This has no dependency at all. Wouldn't that be configured, even if I install something else?

thunder.profile
```diff
@@ -232,18 +232,14 @@ function thunder_themes_installed($theme_list) {
...
-    $ampTheme = $ampThemeConfig->get('amptheme');
-    if (empty($ampTheme) || $ampTheme == 'ampsubtheme_example' || $ampTheme == 'amptheme') {
-      $ampThemeConfig->set('amptheme', 'thunder_amp')
-        ->save(TRUE);
-      \Drupal::configFactory()
-        ->getEditable('amp.settings')
-        ->set('amp_library_process_full_html', 1)
-        ->save(TRUE);
-    }
+    \Drupal::configFactory()
```
goldquest: This can't be solved via optional config?

config/optional/amp.theme.yml
chrfritsch: hm, that's true. if you install amp module without thunder_amp then you will also have this setting. which will be useless

thunder.profile
chrfritsch: no, sadly not

config/optional/amp.theme.yml
goldquest: Do you think, we really need this as an optional config? We could let it be in the themes_installed hook. At least we have everything in the same spot then.

chrfritsch:
- Images in the gallery have different heights and don't respect the focal point
- Instagrams are totally broken, could maybe fixed by using https://www.drupal.org/node/2807735

chrfritsch:
- Under the gallery is a lot of whitespace

goldquest:
I decided to use an existing Image Style. But AMP installs two image styles, which are not using focal point. I think it is ok to leave them there as the "default". What do you think?

goldquest:
There is also the issue, that when you try to install AMP during the thunder installation process, the blocks aren't configured properly.
It would be expected, that most are diabled, but they only are, if the theme gets activated after installation is finished.

goldquest:
We maybe need a warning / requirement check, when installing the AMP theme / module later (Not as optional on installation)

PR #236, opened 2017-02-08
IEF was not saved, destination parameter breaks button

chrfritsch:


modules/thunder_article/src/ThunderNodeForm.php
```diff
@@ -27,6 +27,13 @@ protected function actions(array $form, FormStateInterface $form_state) {
       if ($node->isNew() || $element['unpublish']['#weight'] < $element['publish']['#weight']) {
         $element['save_continue']['#published_status'] = FALSE;
       }
+
+      if ($this->moduleHandler->moduleExists('inline_entity_form')) {
```
mtodor: I think more strict check should be added here. Same as in alter hook of IEF module.
```php
  $widget_state = $form_state->get('inline_entity_form');
  if (!is_null($widget_state)) {
```
In that case callback will be registered only in case of form with IEF and not all content type forms (in our case: Basic Page). I would keep still check for existence of module, just to be on safe side.

PR #242, opened 2017-02-13
Integrate with content_lock

chrfritsch:


thunder.install
```diff
@@ -723,3 +724,40 @@ function thunder_update_8112() {
...
+function thunder_update_8113() {
+
+  /** @var \Drupal\thunder_updater\Updater $thunderUpdater */
+  $thunderUpdater = \Drupal::service('thunder_updater');
+  $updateLogger = $thunderUpdater->logger();
+
+  $modulesInstalledSuccessfully = TRUE;
+
+  try {
+    \Drupal::service('module_installer')->install([
```
mtodor: Would it be possible to use: `Updater::installModules()` here, since we have nice logging there and also error handling. If you need feedback from method is install successful you can add that functionality.

PR #249, opened 2017-02-17
Fixing recursive rebuild issue with thunder_updater

chrfritsch:


chrfritsch:
This fixes one of the two issues we have.

Second one must be fixed in simple_sitemap module. I opened an issue for that: https://www.drupal.org/node/2853459

src/Tests/Installer/ThunderInstallerTest.php
```diff
@@ -162,6 +162,27 @@ public function testInstalled() {
     $this->assertResponse(200);
     // Confirm that we are logged-in after installation.
     $this->assertText($this->rootUser->getUsername());
+
+    /** @var \Drupal\Core\Database\Query\SelectInterface $query */
+    $query = \Drupal::database()->select('watchdog', 'w');
+
+    /** @var \Drupal\Core\Database\Statement $results */
+    $results = $query->fields('w', array(
```
mtodor: I'm not sure, but I think there is no need to fetch all columns (it can be only wid) or even better would be to just execute "count(*)" on database side and compare it in PHP.

PR #261, opened 2017-03-07
Feature/liveblog

goldquest:


src/Plugin/Thunder/OptionalModule/Liveblog.php
```diff
@@ -0,0 +1,74 @@
...
+  /**
+   * {@inheritdoc}
+   */
+  public function buildForm(array $form, FormStateInterface $form_state) {
+
+    $form = parent::buildForm($form, $form_state);
+
+    $form['thunder_liveblog']['description'] = [
+      '#type' => 'item',
+      '#markup' => $this->t('Register a new account at <a href=":pusher_url" target="_blank">:pusher_url</a>, create a new app and note down your keys. You can provide theme right here or at a later stage on the liveblog settings form',
```
chrfritsch: Please mention that the pusher app must be registered in the US cluster. Also add that note to the liveblog settings page.

src/Plugin/Thunder/OptionalModule/Liveblog.php
```diff
@@ -0,0 +1,74 @@
...
+      '#type' => 'textfield',
+      '#title' => t('Key'),
+    ];
+
+    $form['thunder_liveblog']['pusher_secret'] = [
+      '#type' => 'textfield',
+      '#title' => t('Secret'),
+    ];
+
+    $form['thunder_liveblog']['pusher_cluster'] = [
```
chrfritsch: Cluster field should be a select box, like it is on pusher app creation. Also add that to the liveblog settings page

modules/thunder_liveblog/thunder_liveblog.info.yml
```diff
@@ -0,0 +1,17 @@
+name: 'Thunder Liveblog'
```
chrfritsch: Add a "Liveblog" paragraph into thunder_liveblog to make it possible to integrate a liveblog into an article

src/Plugin/Thunder/OptionalModule/Liveblog.php
goldquest: You don't have to register it in the US cluster, but I will add to the description, that you have to note down your cluster.

src/Plugin/Thunder/OptionalModule/Liveblog.php
goldquest: Yeah, I also thought about that. Unfortunately, I can't get the available cluster from the Pusher API, so we would have to track changes there and modify it, each time.

goldquest: I filed a ticket https://support.pusher.com/hc/en-us/requests/10412.
When someone opens up a new app, he also gets the cluster as a string, so I think it is ok for now: 
screenshot 2017-03-08 11 28 17

modules/thunder_liveblog/thunder_liveblog.info.yml
goldquest: hmm, this wasn't planned like that, since the liveblog lazyloads, so it must be always added to the end and it should be easily possible to set the "live" flag.

goldquest: I would propose adding paragraphs to the Liveblog Content Type

goldquest: Or I could add an entity reference to liveblog at the end of an article. The downside of this could be that you can't easily mark your liveblogs in a special way on the frontpage, or make a view with all articles which have a liveblog.

goldquest:
@mtodor Regarding 2) Which header do you mean? From an individual Liveblog Post?

mtodor:
@goldquest Regarding 2) That's why I quoted it there ("header"), because I don't know how to describe it.

2nd try: What is displayed when you don't have any post -> that's liveblog content (static liveblog part). When you click on Edit on liveblog from content list page.

goldquest:
1. is adressed in seperate ticket
   - PR exists
2. is adressed in seperate ticket
3. is fixed
4. fixed / not tagged yet / waiting for (1)
5. is adressed in seperate ticket

mtodor:
From my point of view, this looks good.

PR #262, opened 2017-03-09
Feature/fix apcu install

dbosen:


scripts/travis/02-before-install.sh
```diff
@@ -52,7 +52,10 @@ if [[ $TRAVIS_PHP_VERSION = '5.6' ]] ; then
   yes '' | pecl install yaml;
 elif [[ $TRAVIS_PHP_VERSION = '7.1' ]] ; then
  yes '' | pecl install yaml-2.0.0;
- yes '' | pecl install apcu;
+
+ if ! pecl list | grep apcu >/dev/null 2>&1 ; then
```
mtodor: Maybe we can simply remove this? If apcu is installed and enabled by default by travis for php 7.1, then there is no need to do it explicitly anymore.
Looks like travis did something regarding this: https://github.com/travis-ci/php-src-builder

PR #263, opened 2017-03-10
Update composer.json

landsman:
Blazy - merged patches removed
- https://www.drupal.org/node/2846560
- https://www.drupal.org/node/2849191

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -64,10 +64,6 @@
             "drupal/better_normalizers": {
                 "Add rest dependency": "https://www.drupal.org/files/issues/add_rest_dependency-2849133-8.patch"
             },
-            "drupal/blazy": {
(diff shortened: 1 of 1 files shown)


mtodor:
@landsman Can you please explain reason to remove listed patches?
Because there is problem with removing of them, since current released versions of Blazy is 8.x-1.0-rc1 and that one does not include listed patches.

Additionally to composer "packaging" we have also Drupal packaging (we add dependencies in drupal-org.make.yml -> and then we use premake.sh script to build drupal-org.make from that) and that is used to package distribution on drupal.org and build zip file. List of used modules (including versions and patches) should be same in composer and drupal make packaging files.

mtodor:
We need these 2 patches, so closing pull request.

PR #264, opened 2017-03-13
Move Logger

chrfritsch:

### DIFF
--- a/modules/thunder_updater/thunder_updater.services.yml
+++ b/modules/thunder_updater/thunder_updater.services.yml
@@ -2,3 +2,5 @@ services:
   thunder_updater:
     class: Drupal\thunder_updater\Updater
     arguments: ['@user.shared_tempstore', '@config.factory', '@module_installer', '@current_user']
+  logger.thunder_updater:
+    class: Drupal\thunder_updater\UpdateLogger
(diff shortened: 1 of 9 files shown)


modules/thunder_updater/thunder_updater.services.yml
```diff
@@ -2,3 +2,5 @@ services:
   thunder_updater:
     class: Drupal\thunder_updater\Updater
     arguments: ['@user.shared_tempstore', '@config.factory', '@module_installer', '@current_user']
+  logger.thunder_updater:
```
dbosen: Please rename to thunder_updater.logger

thunder.install
```diff
@@ -314,21 +313,16 @@ function thunder_update_8102() {
...
   $moduleInstaller = \Drupal::service('module_installer');
 
-  try {
-    if (!$moduleInstaller->install(['thunder_updater'])) {
-      $updateLogger->warning(t('Unable to enable the Thunder Updater'));
-    }
-  }
-  catch (MissingDependencyException $e) {
-    $updateLogger->warning(t('Unable to enable the Thunder Updater because of missing dependencies.'));
+  if (!$moduleInstaller->install(['thunder_updater'])) {
```
mtodor: ```ModuleInstaller::install``` will return TRUE or throw (there is no FALSE). So it's possible just to encapsulate everything in try/catch and throw exception -> if you want to provide more descriptive message to user. Or just to let exception from install to bubble up.

Also exceptions should not be translated.

PR #283, opened 2017-04-06
Feature/remove npm dependency

dbosen:


mtodor:
I think we can also remove: package.json - since we don't need bower anymore.
Since we "replace" packages like: jquery, underscore, etc. - should we also replace them for: "npm-asset" vendor?

scripts/travis/04-install-thunder.sh
```diff
@@ -37,7 +37,7 @@ drush_make_thunder() {
 
 composer_create_thunder() {
     cd ${THUNDER_DIST_DIR}
-    composer create-project burdamagazinorg/thunder-project ${TEST_DIR} --stability dev --no-interaction --no-install
+    composer create-project burdamagazinorg/thunder-project:dev-feature/remove-nvm ${TEST_DIR} --stability dev --no-interaction --no-install
```
mtodor: I think we should make new branch + release of thunder-project (8.2.0 I guess) and then use that here instead of branch.

mtodor: I think we should make new branch + release of thunder-project (8.2.0 I guess) and then use that here instead of branch.

composer.json
```diff
@@ -77,6 +64,7 @@
         "burdamagazinorg/infinite_module": "^1.0",
         "burdamagazinorg/infinite_theme": "^1.0",
         "composer/installers": "^1.0",
+        "oomphinc/composer-installers-extender": "^1.1",
```
mtodor: I think that ```oomphinc/composer-installers-extender``` package is not need anymore.

Also I see that we have in composer:
```
    "provide": {
        "heiseonline/shariff": "master"
    },
```
and I'm not sure why it's needed?

PR #289, opened 2017-04-13
Add images to liveblog paragraphs

### LINKED ISSUE #288: Liveblog: add Image paragraph

goldquest:
fixes #288 


goldquest:
Do we need an update hook for this?

modules/thunder_liveblog/thunder_liveblog.install
```diff
@@ -148,3 +150,28 @@ function thunder_liveblog_install() {
...
+ * Adds image to the list of available paragraphs.
+ */
+function thunder_liveblog_update_8101() {
+  /** @var \Drupal\thunder_updater\Updater $thunderUpdater */
+  $thunderUpdater = \Drupal::service('thunder_updater');
+
+  /** @var \Drupal\field\Entity\FieldConfig $field */
+  $field = FieldConfig::load('liveblog_post.liveblog_post.field_embed_media');
+
+  if ($field) {
```
mtodor: Handling should be more failure proof and please add logging for successes + fails. So only if `$field->save()` is successful -> mark checkbox and log success Additionally cases when save() fails or there is no field should be accordingly logged.

tests/src/FunctionalJavascript/ModuleIntegrationTest.php
```diff
@@ -393,4 +395,182 @@ public function testDevicePreview() {
...
+
+    $this->clickDropButton('field_embed_media_image_add_more', FALSE);
+
+    $this->selectMedia("field_embed_media_0_subform_field_image", 'image_browser', ['media:1']);
+
+    $this->liveblogSetBody($page, 'Very nice image post you have here!');
+
+    $this->liveblogSubmit();
+
+    $this->waitUntilVisible('article[data-postid="2"]', 10000);
```
goldquest: But doesn't hurt either, or? This represents, how the content is loaded.

tests/src/FunctionalJavascript/ModuleIntegrationTest.php
```diff
@@ -393,4 +395,182 @@ public function testDevicePreview() {
...
+    $this->waitUntilVisible('article[data-postid="2"] img.b-loaded', 10000);
+
+    // Add post with twitter.
+    $this->liveblogSetTitle($page, 'Twitter post');
+
+    $this->clickDropButton('field_embed_media_twitter_add_more');
+    $this->setFieldValue($page,
```
goldquest: I wanted to have the form filled in from top to bottom. 
It would be possible, to handle all inputs in one call, if we somehow check for ckeditor, but it would still not represent the workflow a user has, when using the module.

tests/src/FunctionalJavascript/ModuleIntegrationTest.php
```diff
@@ -393,4 +395,182 @@ public function testDevicePreview() {
...
+      $page,
+      "textarea[name='body[0][value]']",
+      $body
+    );
+  }
+
+  /**
+   * Submit the liveblog post.
+   */
+  protected function liveblogSubmit() {
```
goldquest: ah nice, thanks

tests/src/FunctionalJavascript/ModuleIntegrationTest.php
```diff
@@ -393,4 +395,182 @@ public function testDevicePreview() {
...
+
+  /**
+   * Click a button within a dropdown button field.
+   *
+   * @param string $fieldName
+   *   The [name] attribute of the button to be clicked.
+   * @param bool $toggle
+   *   Whether the dropdown button should be expanded before clicking.
+   */
+  protected function clickDropButton($fieldName, $toggle = TRUE) {
```
goldquest: Unfortunately not, since the `addParagraph` method needs the "addText" paragraph, which is not present in the liveblog post form.

Either I rewrite the `addParagraph` method to be more generalised, or I move the `clickDropButton` to the Base class.

modules/thunder_liveblog/thunder_liveblog.install
goldquest: will do

goldquest: `field->save` actually throws an error, if something went wrong. Do you want me to catch it and log accordingly, instead of letting it bubble up?

goldquest: BTW, this is also true for other tests, e.g. in the `thunder.install` file. Maybe we should adjust them.

tests/src/FunctionalJavascript/ModuleIntegrationTest.php
mtodor: It should not depend on any type of paragraph -> but I have forgot that we have switched to modal dialog for adding paragraphs. In your case you can't use that method.

So just make your function generic (if it's not already) and move to base.

modules/thunder_liveblog/thunder_liveblog.install
mtodor: If we let it bubble -> it could block all hooks after this one. So I would rather catch it and log it, check point will stay unchecked and user have to check what is wrong manually anyway. If you find any 81xx update hook that doesn't handle errors properly, please create Trello card for it and tag it accordingly.

tests/src/FunctionalJavascript/ModuleIntegrationTest.php
mtodor: It's ok to keep it like that. It doesn't hurt, that's true. And it could also be helpful for debugging, if 1st pass, but 2nd fails. It's only, that I'm not fan of waiting .. :)

tests/src/FunctionalJavascript/ModuleIntegrationTest.php
mtodor: Yes, since paragraphs are between title and body of Live Blog. That's true. Good point. 👍 
I was thinking mainly to get rid of one line wrapping functions, that's all.

modules/thunder_liveblog/thunder_liveblog.install
goldquest: Added a trello card

tests/src/FunctionalJavascript/ModuleIntegrationTest.php
goldquest: :) they should be executed pretty fast. The longer waiting time is only added, if resources are slow on travis.

PR #301, opened 2017-04-28
[WIP]: Add pluggable access control handler

stborchert:
Replaces the AccessControlHandler provided by _thunder_taxonomy_ with a pluggable version for Nodes, Comments and TaxonomyTerms.


dbosen:
@stborchert: Could you please explain the reasoning behind this? It really does not look like something, that I want in a distribution.

stborchert:
@dbosen: The main reason for this is the access structure needed by thunder-community (forum access management). Overriding each AccessControlHandler is not an option since Thunder itself (_thunder_taxonomy_) is doing this and you can't have multiple handlers for the same entity type. So its best to make this pluggable and provide a bunch of plugins for each use case (i.e. Term status, Node access within forum terms, ...).

I agree with you that this shouldn't be done in a distribution. Ideally Drupal itself would provide  a pluggable access control system but ... we don't wan't to wait for Drupal 9 ;)

stborchert:
The changes are not necessary anymore as the pluggable access control handler will be provided by a contributed module.

PR #315, opened 2017-05-16
AMP integration tests

goldquest:


tests/src/FunctionalJavascript/ModuleIntegrationTest.php
```diff
@@ -542,4 +542,67 @@ protected function liveblogSetBody(DocumentElement $page, $body) {
...
+
+    // Image paragraph.
+    $this->assertSession()->elementExists('css', '.paragraph--type--image amp-img');
+    $this->assertSession()->waitForElementVisible('css', '.paragraph--type--image amp-img img');
+
+    // Gallery paragraph.
+    $this->assertSession()->elementExists('css', '.paragraph--type--gallery amp-carousel');
+    // Images in gallery paragraph.
+    $this->assertSession()->elementsCount('css', '.paragraph--type--gallery amp-carousel amp-img', 3);
+    $this->assertSession()->waitForElementVisible('css', '.paragraph--type--gallery amp-carousel amp-img');
```
mtodor: Why is this wait here, since it's already asserted that elements exist? They are not visible in moment of count assertion or?

tests/src/FunctionalJavascript/ModuleIntegrationTest.php
```diff
@@ -542,4 +542,67 @@ protected function liveblogSetBody(DocumentElement $page, $body) {
...
+
+    // Link paragraph.
+    $this->assertSession()->linkExists('Google');
```
mtodor: Can you also assert link: `$this->assertSession()->linkByHrefExists()`

tests/src/FunctionalJavascript/ModuleIntegrationTest.php
goldquest: Yes, it should first be waited and then be counted

tests/src/FunctionalJavascript/ModuleIntegrationTest.php
goldquest: sure

composer.json
```diff
@@ -78,6 +78,12 @@
             },
             "drupal/paragraphs": {
                 "create a NEW paragraph 'in place' between existing ones": "https://www.drupal.org/files/issues/sorting_of_paragraphs-2828110-6.patch"
+            },
+            "drupal/amp": {
```
goldquest: sure

tests/src/FunctionalJavascript/ModuleIntegrationTest.php
goldquest: actually, I removed it, because we just use existing articles.

tests/src/FunctionalJavascript/ModuleIntegrationTest.php
```diff
@@ -542,4 +542,67 @@ protected function liveblogSetBody(DocumentElement $page, $body) {
     );
   }
 
+  /**
+   * Testing integration of "AMP" module and theme.
+   */
+  public function testAmpIntegration() {
```
mtodor: My plan is to have module with test content and switch tests to use it. It will probably be based on demo content in first hand, but later it will be easier to add when something is needed.

tests/src/FunctionalJavascript/ModuleIntegrationTest.php
```diff
@@ -542,4 +542,44 @@ protected function liveblogSetBody(DocumentElement $page, $body) {
...
```
dbosen: Is it possible to  actually test for a valid AMP page?

PR #339, opened 2017-06-22
Update thunder.install

zuernBernhard:
Updated failed here because we had no riddle media


mtodor:
@zuernBernhard - can you explain problem? How it manifests?
Because media based riddles should exist, since it's provided with "thunder_riddle" module.

Btw. I'll close other PR #340 - since it's duplicate.

zuernBernhard:
Yes, you are right we can close this - I could fix the problem with an update hook.

mtodor:
@zuernBernhard anyway it would be better to switch to media based riddles, because then you can use entity browser provided for that. It's way nicer experience, since you get full list of riddles with thumbnails provided by Riddle.

PR #352, opened 2017-07-12
Fix media paragraph style

dbosen:


modules/thunder_paragraphs/thunder_paragraphs.install
```diff
@@ -101,3 +101,29 @@ function thunder_paragraphs_update_8001() {
...
+
+/**
+ * Change media paragraph to use inline_entity_form.
+ */
+function thunder_paragraphs_update_8101() {
+  /** @var \Drupal\thunder_updater\Updater $thunderUpdater */
+  $thunderUpdater = \Drupal::service('thunder_updater');
+  $updateLogger = $thunderUpdater->logger();
+
+  $config = \Drupal::configFactory()->getEditable('core.entity_form_display.paragraph.media.default');
```
mtodor: It would be best to make 3 cases here:
1. we should check if media paragraph exists on system at all and if not -> we check `v1_1__media_paragraph` in CheckList and log that nothing is changed and everything is good (so `info()` log)
2. if media paragraph exists + we execute update successfully -> we mark update as successful, but with unchecked `v1_1__media_paragraph`, so that we ensure admin will ready notification of deprecation
3. if media paragraph exists + update fails, because field widget is not "rendered_entity" -> then we mark update as failed

modules/thunder_updater/updates.yml
```diff
@@ -90,3 +90,7 @@ v1.1:
         '#title': 'New Thunder Admin UX'
         '#description': "<p>Thunder now comes with new improved administration user interface.</p>"
         '#description_failed': '<p><strong>There was a problem during the update.</strong> New Thunder Admin UX require plenty new configurations to be adjusted and imported in order to use it properly. We suggest you to take a look at execution of update hook for enabling new administration theme and manually adjust missing configurations.<p>'
+    v1_1__media_paragraph:
+        '#title': 'Updated media paragraph'
+        '#description': "<p>Last update of media paragraph to be compatible with thunder admin theme. The media paragraph is now considered deprecated and will not get any updates in the future.</p>"
```
mtodor: Should we emphasize (with <strong\> or so) that it's deprecated. Also text should be adjusted for "failed" case if we change logic in update hook.

PR #376, opened 2017-08-07
Issue #2899917 by mtodor: Paragraphs add in between buttons styling is not good

mtodor:
Using of new paragraphs add in between buttons patch.


composer.json
```diff
@@ -135,7 +135,7 @@
         "drupal/shariff": "^1.0",
         "drupal/slick": "^1.0",
         "drupal/slick_media": "^1.0",
-        "drupal/thunder_admin": ">=1",
+        "drupal/thunder_admin": "dev-feature/2899917-add-in-between-buttons-styling",
```
dbosen: This has to be changed back to ">=1"

PR #403, opened 2017-09-21
fixed typo;

landsman:


mtodor:
@landsman Thank you for contribution.

It's a bit difficult for us to work with external pull requests (no rights, limited tests, etc.), that's why I have pulled your commit into other branch (PR #404, commit stays authored by you), so that I can easier test everything with UPDATE_TEST path (that means tests will execute updates and test Thunder after that).

So this PR can be closed, since everything will be addressed in other pull request.

PR #404, opened 2017-09-22
Issues #2910627 and #2910773 by landsman, mtodor: Typo fix in thunder_updater and fixing of responsive_preview integration tests

mtodor:


mtodor:
Looks like problem is caused by chrome update. More info: https://stackoverflow.com/questions/45061901/chrome-61-body-doesnt-scroll

PR #409, opened 2017-09-27
Issue #2908456 by mtodor: Live blog tests flip flop

mtodor:


mtodor:
First commit is just export of test to separate Class. After that changes can be reviewed.

Following things has changed:
```
$this->waitUntilVisible('[data-tweet-id="778001033142284288"].twitter-tweet-rendered', 10000);

// We can't check inside Twitter widget is it loaded or not, that's why
// plain wait is used.
$this->getSession()->wait(5000);
```

I run solution with plain wait and here are results with number of errors:
```
PHP 7.1 - 0/30
PHP 5.6 - 0/50
```

And solution without plain wait and here are results with number of errors:
```
PHP 7.1 - 0/30
PHP 5.6 - 2/10
```

Looks like, problem is that it takes a lot time for Twitter to load - and it's not possible to check it in good way.

PR #410, opened 2017-09-28
Prepare Thunder 3.x Branch

chrfritsch:
Lets see how this will go


drupal-org.make
```diff
@@ -68,20 +68,17 @@ projects[linkit][version] = 4
...
+projects[media_entity_slideshow][version] = 2
 projects[media_entity_twitter][type] = module
-projects[media_entity_twitter][version] = 1
+projects[media_entity_twitter][version] = 2
 projects[media_expire][type] = module
-projects[media_expire][version] = 1
+projects[media_expire][version] = 2
 projects[metatag][type] = module
 projects[metatag][version] = 1
 projects[nexx_integration][type] = module
```
chrfritsch: The nexx module would not work in that version

chrfritsch (commented):
Beside that nexx thing, I would say let's go

PR #412, opened 2017-09-28
Issue #2912475: Update paragraphs to version ^8.x-1.2

C-Logemann:
Here is my related issue on d.o: https://www.drupal.org/node/2912475

Hi I'm just starting with thunder development and have problems with the suggested dev and testing setup but this is another issue. So I currently can't run the tests. But I believe this is just a minor change.


chrfritsch:
There is much more stuff to do, to update to the latest paragraphs version. We will handle everything in https://github.com/BurdaMagazinOrg/thunder-distribution/pull/413, because we want to switch to the experimental widget. I will close this PR

PR #413, opened 2017-10-02
Issue #2908887: Move to Paragraphs experimental Widget

chrfritsch:


dubcanada:
What's the status of this? I'd like to use the experimental widget. Does anything need to be finished/tested?

dbosen (changes requested):
~~The option to confirm deletion of paragraphs should be enabled by default for both the article and the term paragraphs field.~~

mtodor:
~~We should add testes for - delete confirmation and single option to button features on paragraphs. That could be in one test, both tests are related to remove action.~~

modules/thunder_updater/updates.yml
```diff
@@ -209,3 +209,10 @@ v2.16:
         '#description_successful': '<p>The installation was successful.</p>'
         '#description_failed': '<p><strong>Installation failed</strong><p>'
         thunder_article_config: { '#text': 'Move scheduler link back from local tasks to menu tab', '#url': 'internal:/admin/config/thunder_article/configuration' }
+v2.17:
```
mtodor: This should be adjusted, because version has changed.

scripts/travis/04-install-thunder.sh
```diff
@@ -38,7 +38,7 @@ drush_make_thunder() {
 
     # Get development branch of Thunder Admin theme (to use same admin theme as for composer build)
     rm -rf ${TEST_DIR}/docroot/profiles/thunder/themes/thunder_admin
-    git clone --depth 1 --branch 8.x-2.x https://git.drupal.org/project/thunder_admin.git ${TEST_DIR}/docroot/profiles/thunder/themes/thunder_admin
+    git clone --depth 1 --single-branch --branch "feature/2908887-experimental-paragraphs-widget" https://github.com/BurdaMagazinOrg/theme-thunder-admin.git ${TEST_DIR}/docroot/profiles/thunder/themes/thunder_admin
```
mtodor: ~~Need to be adjusted before merge and we need release of theme first anyway.~~

tests/src/FunctionalJavascript/ThunderJavascriptTestBase.php
```diff
@@ -112,6 +112,20 @@ protected function getDriverArgs() {
...
   }
 
+  /**
+   * Get WebDriver URL, that can be set by environment variable.
+   *
+   * @return string
+   *   Returns full URL to WebDriver interface.
+   */
+  protected function getWebDriverUrl() {
+    if (!empty(getenv('THUNDER_WEBDRIVER_HOST'))) {
```
mtodor, line 122: ~~Should be documented in developer documentation.~~

mtodor (changes requested):

composer.json
```diff
@@ -124,7 +120,8 @@
         "drupal/media_expire": "^1.0",
         "drupal/nexx_integration": "^1.0||^2.0",
         "drupal/metatag": "^1.0",
-        "drupal/paragraphs": "1.1",
+        "drupal/paragraphs": "^1.0",
+        "drupal/paragraphs_features": "dev-1.x",
```
dbosen: we need to wait for paragraphs features release, and change this line before merge

composer.json
```diff
@@ -134,7 +131,7 @@
         "drupal/shariff": "^1.0",
         "drupal/slick": "^1.0",
         "drupal/slick_media": "^1.0",
-        "drupal/thunder_admin": ">=1",
+        "drupal/thunder_admin": "dev-2.x",
```
dbosen: ~~Has to be reverted before merge~~

thunder.post_update.php
```diff
@@ -4,3 +4,138 @@
...
+  }
+
+  // Update is successful only when all steps are successful.
+  $successful = TRUE;
+
+  // There are few new modules required for experimental paragraphs widget.
+  if (
+    !$thunderUpdater->installModules(['thunder__thunder_update_8120' => 'default_content'])
+    || !$thunderUpdater->installModules(['thunder__thunder_update_8120' => 'better_normalizers'])
+    || !$thunderUpdater->installModules(['thunder__thunder_update_8120' => 'paragraphs_features'])
```
dbosen: Module installation should be done in update hook

thunder.install
```diff
@@ -951,3 +951,138 @@ function thunder_update_8120() {
...
+  // Update is successful only when all steps are successful.
+  $successful = TRUE;
+
+  // There are few new modules required for experimental paragraphs widget.
+  if (
+    !$thunderUpdater->installModules(['thunder__thunder_update_8121' => 'default_content'])
+    || !$thunderUpdater->installModules(['thunder__thunder_update_8121' => 'better_normalizers'])
+    || !$thunderUpdater->installModules(['thunder__thunder_update_8121' => 'paragraphs_features'])
+  ) {
+    $successful = FALSE;
```
dbosen: This check leads to the problem, that if either default_content or better_normalizers is missing, the whole update fails. Since we would still be able to do most of the updates, this should be changed.
Proposal: 
Split update hook into multiple update hooks, each of them with different dependencies that define if the update was successful. This way we can do most updates, that for example only need the paragraphs feature modules, and let only the update fail, that actually needs default_content or better_normalizers. Also, the Liveblog update could be separated out and only done when needed.
The general idea is to separate those updates that have independent requirements.

chrfritsch:
~~So I am testing things a bit. This one might be out of scope, but I wanted to note it anyway :-)~~

~~If the "Autocollapse" feature is enabled and you change something in an image paragraph, then switch to a text pragraph, your changes are lost :-(~~

chrfritsch:
~~The "Add above" functionality doesn't work properly when "Add in-between" is enabled. I tested that in Seven theme, because "Add above" doesn't work in "Thunder admin" currently.~~

mtodor (approved):

dbosen (approved):

PR #416, opened 2017-10-04
Issue #2899242: Update Thunder to Drupal 8.4

chrfritsch:


tests/src/FunctionalJavascript/ThunderJavascriptTestBase.php
```diff
@@ -453,24 +453,38 @@ public function runCron() {
...
+    $page->find('xpath', '//div[@data-drupal-selector="edit-actions"]/input[@id="edit-submit"]')
+      ->click();
+  }
+
+  /**
+   * Set entity status.
    *
-   * 1 - Save as unpublished (default).
-   * 2 - Save and publish.
+   * TRUE - Unpublished.
```
mtodor: Comment should be adjusted - it's a bit confusing.

And maybe it would be good to rename method, so it's clear that Published status will be set - maybe: `setPublishedStatus` or something like that.

maosmurf:
Our only issue with current stable thunder (8.2.8) and drupal 8.4 was `Yaml::parse` in `thunder_updater.module`.

As we didn't want to wait for full release in order to test 8.4 with thunder, we patched `"burdamagazinorg/thunder"` 
[Parse_content_of_updates_yml.patch.zip](https://github.com/BurdaMagazinOrg/thunder-distribution/files/1373452/Parse_content_of_updates_yml.patch.zip)

PR #434, opened 2017-11-07
Issue #2899254: Integrate editorial search with Search API module

chrfritsch:


modules/thunder_search/config/optional/facets.facet.content_type.yml
```diff
@@ -0,0 +1,60 @@
...
+    show_numbers: false
+    default_option_label: Choose
+query_operator: or
+use_hierarchy: false
+expand_hierarchy: false
+enable_parent_when_child_gets_disabled: true
+hard_limit: 0
+exclude: false
+only_visible_when_facet_source_is_visible: true
+processor_configs:
```
dbosen: hide_1_result_facet is missing for content type

dbosen (changes requested):
Tag based caching breaks second request to contetn view

dbosen (changes requested):
Several actions are missing on the content list

dbosen (changes requested):
Please move the "Fulltext search" box to the front of the form

dbosen (changes requested):
Also add search to media tab and media browser

dbosen (changes requested):
Tests are missing!

dbosen (changes requested):
Please add channel and tags to facets.

dbosen (changes requested):
When the list gets filtered, a " Selected X items in this view" box gets added to the form. This seems to be very buggy.
When selecting Items, they get correctly counted, but the titles are not shown. 
When no Selection has been made outside of the current scope, the box should not be shown.
Deselecting Items does not work, they keep beeing selected.

dbosen (changes requested):
Add author to filter or facette

modules/thunder_search/config/optional/views.view.media_search_api.yml
```diff
@@ -0,0 +1,887 @@
...
+  module:
+    - datetime
+    - image
+    - search_api
+    - thunder_search
+    - user
+    - views_bulk_operations
+    - config_selector
+id: media_search_api
+label: Media
```
mtodor: Label for Media view with search API should be adjusted or at least Description. Since in Configuration Selector there are 2 views with same name and no description. :(

dbosen:
The media view has two status facettes
bildschirmfoto 2018-04-23 um 10 58 47

dbosen:
Remove promoted and sticky facettes.
bildschirmfoto 2018-04-23 um 11 28 25

dbosen:
Author and channel should be autocomplete widgets.
When autocomplete widget is available, we could also add tha tags as a facette

chrfritsch:
Known issue: https://www.drupal.org/project/views_bulk_operations/issues/3009171

modules/thunder_search/src/Plugin/views/area/FacetBlock.php
```diff
@@ -0,0 +1,70 @@
...
+  /**
+   * {@inheritdoc}
+   */
+  public function render($empty = FALSE) {
+
+    /** @var \Drupal\facets\FacetManager\DefaultFacetManager $facetManager */
+    $facetManager = \Drupal::service('facets.manager');
+
+    /** @var \Drupal\facets\FacetSource\FacetSourcePluginManager $source */
+    $sources = \Drupal::service('plugin.manager.facets.facet_source');
```
dbosen: Inject it please!

tests/src/Functional/ModuleUninstallTest.php
```diff
@@ -35,6 +35,7 @@ class ModuleUninstallTest extends ThunderTestBase {
     ['length_indicator'],
     ['redirect'],
     ['simple_sitemap'],
+    ['thunder_search', 'search_api_db', 'search_api', 'facets'],
```
mtodor: I'm not 100% sure, but should we add all enabled modules to list. This is list of modules enabled:
```
thunder_search
views_bulk_operations
select2_facets
search_api_db
search_api
facets
```

thunder.post_update.php
```diff
@@ -0,0 +1,37 @@
...
+  if ($thunderUpdater->executeUpdates([['thunder', 'thunder_add_config_selector_to_content_media']])) {
+    if ($view = View::load('content')) {
+      $view->save();
+    }
+    if ($view = View::load('thunder_media')) {
+      $view->save();
+    }
+    /** @var \Drupal\config_update\ConfigReverter $configReverter */
+    $configReverter = \Drupal::service('config_update.config_update');
+    $configReverter->import('config_selector_feature', 'thunder_content_view');
```
mtodor: Maybe we should wrap imports into try/catch if for some reason import fails.

modules/thunder_search/src/Plugin/views/area/FacetBlock.php
```diff
@@ -0,0 +1,70 @@
...
+    $build = ['#type' => 'container', '#attributes' => ['class' => 'form--inline clearfix']];
+    $build['_view'] = $this->view->display_handler->viewExposedFormBlocks();
+    foreach ($facetManager->getFacetsByFacetSourceId($facet_source->pluginId) as $id => $facet) {
+      // No need to build the facet if it does not need to be visible.
+      if (($facet->getOnlyVisibleWhenFacetSourceIsVisible() && !$facet->getFacetSource()->isRenderedInCurrentRequest())) {
+        continue;
+      }
+      $config = $facet->getFacetSource()->getDisplay()->getPluginDefinition();
+
+      if ($config['view_id'] == $this->view->id() && $config['view_display'] == $this->view->current_display) {
```
mtodor: [cleaner code] Since you have already started with checking for wrong values and `continue` with looping, you could do the same for these two `if`s too. That will reduce nesting too.

tests/src/FunctionalJavascript/Integration/ConfigSelectorTest.php
```diff
@@ -0,0 +1,52 @@
...
+
+    $assert_session = $this->assertSession();
+
+    // Content lock fields are there by default.
+    $this->drupalGet('admin/content');
+    $assert_session->elementExists('xpath', '//*[@id="view-title-table-column"]/a');
+    $assert_session->elementExists('css', '#block-thunder-admin-content > div > div.view-content');
+
+    // Install thunder_search.
+    \Drupal::service('module_installer')->install(['thunder_search']);
```
mtodor: [code optimization] Maybe you can assign `\Drupal::service('module_installer')` to variable and use it, since it's used more then once.

mtodor (changes requested):
Here are few suggestions from me:

1.) There should be some loading indicator when I change filtering value, the reason for that is that I (as a user) do not see what's going on and I could trigger new action again.
When I do some changes quickly, for example:
1.a. When I select published and then quickly also promoted
1.b. I will get filtered only by promoted option and published selection is gone
-> Same problem is with removing filters, what could happen way faster
If we would use a mask, then something like that would not be possible. But it's also not so nice solution.

2.) When my search does not match anything, the search form is gone. :(
How to reproduce:
2.a. Enter some word that will not match anything, for example: SomeNoneMatchingText
2.b. You should get empty list and search field is gone

3.) Some characters make JS error
How to reproduce:
3.a. Try to search with following string: some chars are pro'blematic
-> Good side is that search will actually display the correct result

4.) Channel facet is always displayed
When I search for something and there is only one result, Channel facet is still displayed.
From @dbosen - there is a general UX question: Do we want facets to disappear or to disable them? I guess this question is for @klucic.

5.) Search Index fields
I'm not sure is it relevant to have a search by Teaser Text and/or SEO Title.
Idea from @ol0lll - maybe we can use custom view mode for full-text search.

modules/thunder_search/config/optional/search_api.index.content.yml
```diff
@@ -0,0 +1,316 @@
...
+    - label
+    - media_author
+    - name
+    - rendered_item
+    - type
+    weights:
+      preprocess_index: -20
+      preprocess_query: -20
+  rendered_item: {  }
+  stopwords:
```
dbosen: Stopwords should be handled by a real server, not by the search api => disable it

modules/thunder_search/config/optional/search_api.index.content.yml
```diff
@@ -0,0 +1,316 @@
...
+    fields:
+    - label
+    - rendered_item
+    spaces: ''
+    overlap_cjk: 1
+    minimum_word_size: '3'
+    weights:
+      preprocess_index: -6
+      preprocess_query: -6
+  transliteration:
```
dbosen: Transliteration should be disabled for editorial searches

dbosen (changes requested):

chrfritsch:
@mtodor Nr. 3 https://www.drupal.org/project/drupal/issues/2902769

chrfritsch:
@mtodor Regarding 1.) I think this is a general problem with facets, not specifically to select2. So I would propose to build this kind of functionality into the facets module. 
I would also argue that this is a followup and shouldn't block this PR.

mtodor:
Here are some additional findings:

1. If I'm logged as Editor, when I search to some value without result, I'm getting following message: "It's needed to index the search index in order to have all content searchable." - with link to indexing page. And when I click on that link, I get 403. Is it possible to hide this message for users without rights or to totally hide it.

2. There is one strange behaviour, that when I start to narrow down search with facets and then search with text, then everything is reseted. I think that UX could be better with different layout of "filter" form.

modules/thunder_search/thunder_search.info.yml
```diff
@@ -0,0 +1,10 @@
+name: 'Thunder Search'
+description: 'Provides better search integration for Thunder.'
+type: module
+core: 8.x
+dependencies:
+  - facets:facets
+  - search_api:search_api_db
```
mtodor: We should remove dependency to Search API Database module. It's OK to install it with update if it's needed, but not force sites to have that module enabled.

mtodor (changes requested):

chrfritsch:
Also important: https://www.drupal.org/project/search_api/issues/3013663

mtodor:
I have found 2 things:
1. Media view does not work. Every search gives empty result for me.
2. Filters (facets) are not responsive. Someone with smaller screen has to scroll left-right.

When it comes to performance. I tested with 3k articles and it looks like this:
In compare with old content view, Search API view is around 30% slower for Solr and around 150% for Database search server. So **users without Solr** service will not be so happy with this feature.

mtodor:
1. I'm not able to evaluate indexing weights defined for fields. It looks for me that, teaser text weight is too high. To test that properly, we need real data, not generated one.  :(
2. I already suggested that before, can we add loading indicator when the facet is changed? Reducing results with facets can lead to super bad experience.
3. Caching for search results is not working good. Here are some examples:
   - Make 2 articles with one word in Teaser text, for example, "Kupaianaha", then filter the content list by that word. After that edit one article and change Teaser text, that it doesn't contain "Kupaianaha" after you save that article you will be automatically redirected to the content list page with a previously defined filter. And you will have the changed article in the list.
   - I have found another example, easier to reproduce. Filter articles by one Tag -> click edit for the first article -> remove Tag that you have searched from that article -> click Save -> you will get back to a same filtered list.
   - If you filter like this: by author and unpublished -> and then you click edit and publish one article -> you will get redirected back to content list and article will still be in a list of unpublished. Super confusing. :(
   My guess is that Solr doesn't get updated fast enough before the new list is generated. So it's some kind of race condition problem.
4. Related to 3. - I think that cache invalidation does not work properly.

Here is some performance testing of Search API integration.
I want also to test with 15k, need to create articles and then I can test, just to see if performance is scaling with an increase of articles.

| 3k Articles    | Local no Load     |  | Solr Search         | 29,14%     |
|----------------|-------------------|--|----------|------------|
| Search View [ms]    | Normal [ms]           |  | Slowdown [ms] | Slowdown % |
| 1900           | 1500              |  | 400      | 26,67%     |
| 568            | 417               |  | 151      | 36,21%     |
| 645            | 621               |  | 24       | 3,86%      |
| 249            | 207               |  | 42       | 20,29%     |
| 822            | 728               |  | 94       | 12,91%     |
| 740            | 506               |  | 234      | 46,25%     |
| 548            | 400               |  | 148      | 37,00%     |
| 1830           | 1470              |  | 360      | 24,49%     |
| 540            | 400               |  | 140      | 35,00%     |
| 1790           | 1350              |  | 440      | 32,59%     |
| 335            | 220               |  | 115      | 52,27%     |
| 724            | 534               |  | 190      | 35,58%     |
| 309            | 267               |  | 42       | 15,73%     |

| 3k Articles    | Local no Load     |  |  Database Search        | 144,97%    |
|----------------|-------------------|--|----------|------------|
| Search View [ms]    | Normal [ms]           |  | Slowdown [ms] | Slowdown % |
| 556            | 394               |  | 162      | 41,12%     |
| 7950           | 1430              |  | 6520     | 455,94%    |
| 2430           | 1030              |  | 1400     | 135,92%    |
| 2460           | 669               |  | 1791     | 267,71%    |
| 624            | 284               |  | 340      | 119,72%    |
| 1600           | 551               |  | 1049     | 190,38%    |
| 1860           | 1410              |  | 450      | 31,91%     |
| 929            | 402               |  | 527      | 131,09%    |
| 676            | 545               |  | 131      | 24,04%     |
| 603            | 397               |  | 206      | 51,89%     |

| 10k Articles   | AWS on Heavy Load |  | Solr Search         | 49,19%     |
|----------------|-------------------|--|----------|------------|
| Search View [ms]    | Normal [ms]           |  | Slowdown [ms] | Slowdown % |
| 400            | 240               |  | 160      | 66,67%     |
| 1540           | 1120              |  | 420      | 37,50%     |
| 1450           | 891               |  | 559      | 62,74%     |
| 471            | 193               |  | 278      | 144,04%    |
| 776            | 611               |  | 165      | 27,00%     |
| 260            | 176               |  | 84       | 47,73%     |
| 807            | 746               |  | 61       | 8,18%      |
| 442            | 267               |  | 175      | 65,54%     |
| 396            | 211               |  | 185      | 87,68%     |
| 650            | 522               |  | 128      | 24,52%     |
| 635            | 539               |  | 96       | 17,81%     |
| 1660           | 1020              |  | 640      | 62,75%     |
| 1630           | 1080              |  | 550      | 50,93%     |
| 520            | 482               |  | 38       | 7,88%      |
| 330            | 224               |  | 106      | 47,32%     |
| 599            | 448               |  | 151      | 33,71%     |
| 1760           | 1220              |  | 540      | 44,26%     |

| 10k Articles   | AWS no Load       |  | Solr Search        | 33,79%     |
|----------------|-------------------|--|----------|------------|
| Search View [ms]    | Normal [ms]           |  | Slowdown [ms] | Slowdown % |
| 1330           | 1030              |  | 300      | 29,13%     |
| 1150           | 891               |  | 259      | 29,07%     |
| 340            | 217               |  | 123      | 56,68%     |
| 273            | 227               |  | 46       | 20,26%     |

mtodor:
15k Tests

| 15k Articles | AWS no Load |  | Solr Search         | 60,39%     |
|--------------|-------------|--|----------|------------|
| Search View [ms]  | Normal [ms]      |  | Slowdown [ms] | Slowdown % |
| 200          | 130         |  | 70       | 53,85%     |
| 300          | 237         |  | 63       | 26,58%     |
| 963          | 271         |  | 692      | 255,35%    |
| 202          | 142         |  | 60       | 42,25%     |
| 961          | 736         |  | 225      | 30,57%     |
| 1150         | 766         |  | 384      | 50,13%     |
| 277          | 196         |  | 81       | 41,33%     |
| 208          | 142         |  | 66       | 46,48%     |
| 337          | 249         |  | 88       | 35,34%     |
| 438          | 400         |  | 38       | 9,50%      |
| 472          | 273         |  | 199      | 72,89%     |

ol0lll:
@mtodor:

**3 and 4**
Cache invalidation is not perfect but not the problem, unfortunately, I added entity tags to the query here: [3034996](https://www.drupal.org/project/search_api/issues/3034996)

Results are out of date because of the asynchronous design of search_api. Rebuilding the index is done after the response is sent and before the service is terminated, the next request (destination) is therefore served from stale index.

PR #443, opened 2017-11-14
Issue #2931731 by tjwelde: [UX] Field length indicator without setting a hard limit

goldquest:


chrfritsch:
Thats pretty much code and i'm glad that it's not coupled to Thunder. So please create a d.o. project for that.

modules/thunder_article/modules/thunder_indicator/thunder_indicator.module
```diff
@@ -0,0 +1,206 @@
...
+ *   The optimum minimum.
+ * @param int $optimax
+ *   The optimum maximum.
+ * @param int $tolerance
+ *   The tolerance.
+ *
+ * @return array
+ *   The widths and positions of the indicators.
+ */
+function _thunder_indicator_get_width_and_pos($optimin, $optimax, $tolerance) {
```
chrfritsch: I think this should be moved into a service

mtodor:
I have tested this a bit, looks good from my point of view. There are 2 things that I'm not sure of:

1. default numbers for teaser - they are really small and if you base your teaser content on length indicator then teaser doesn't look nice. So it's question, do we kick length indicator from teaser, or change numbers to that teaser looks nice in front-end, or leave it as it is in PR.
2. (length indicator related, but not so relevant) maybe it would be nice to add "change" handler too, because jQuery practice is after setting value for input field -> trigger change event: https://api.jquery.com/val/

goldquest:
Regarding 1: Right now the page description is set to be the field teaser text. Do we maybe need another SEO description field?
Regarding 2: I created an issue for that.

dbosen:
I am also in favor of rethinking the description/teaser field. The length indicator is not working for the description.

PR #450, opened 2017-11-17
Issue #2924323: Make update generation more generic for thunder_updater

chrfritsch:


modules/thunder_updater/thunder_updater.drush.inc
```diff
@@ -34,20 +39,37 @@ function thunder_updater_drush_command() {
...
+    $modules = array_keys($modules);
+  }
+  if ($regex = drush_get_option('filter', '')) {
+    $modules = array_filter($modules, function ($module) use ($regex) {
+      return preg_match($regex, $module);
+    });
+  }
+
+  $patchFile = $configHandler->generatePatchFile($modules);
+  file_put_contents($patchFilePath, $patchFile, FILE_CREATE_DIRECTORY);
```
mtodor: I think this is not correct, since: FILE_CREATE_DIRECTORY - is provided by Drupal and should not be used with: `file_put_contents`.

Additionally, file should not be created if `$patchFile` is empty.

modules/thunder_updater/thunder_updater.drush.inc
```diff
@@ -14,13 +16,16 @@ function thunder_updater_drush_command() {
...
+    ],
+    'options' => [
+      'filter' => 'Regex to filter the module list',
     ],
     'drupal dependencies' => ['thunder_updater'],
-    'required-arguments' => 2,
+    'required-arguments' => 1,
     'examples' => [
-      'drush thunder-updater-generate-update thunder_media thunder_media__update_8099' => 'Generate patch to update configuration from current installed configuration to configuration provided in files.',
+      'drush thunder-updater-generate-update mypatchfile.yml --filter=/thunder/' => 'Generate patch to update configuration from current installed configuration to configuration provided in files.',
```
mtodor: Can you also add 2nd example with usage of `module` argument. Also it would be nice to rename argument to `modules`, fe.

modules/thunder_updater/src/Command/GenerateThunderUpdateCommand.php
```diff
@@ -0,0 +1,314 @@
...
+use Symfony\Component\Console\Output\OutputInterface;
+use Drupal\Console\Core\Command\Command;
+use Drupal\Console\Core\Style\DrupalStyle;
+
+/**
+ * Generate Thunder update command class.
+ *
+ * TODO: Add support for version option, where entry in checklist will be added.
+ *
+ * @Drupal\Console\Annotations\DrupalCommand (
```
chrfritsch: Is the annotation needed? Didn't found anything about that in the docs...
https://hechoendrupal.gitbooks.io/drupal-console/content/en/extending/creating-custom-commands.html

modules/thunder_updater/src/Command/GenerateThunderUpdateCommand.php
```diff
@@ -0,0 +1,314 @@
...
+    }
+  }
+
+  /**
+   * Get update gook generator.
+   *
+   * @return \Drupal\thunder_updater\Generator\ThunderUpdateGenerator
+   *   Update hook generator for Thunder update.
+   */
+  protected function createGenerator() {
```
chrfritsch: This is not needed. You already injected ThunderUpdateGenerator into this class

modules/thunder_updater/src/Command/GenerateThunderUpdateCommand.php
mtodor: I have used `generate:command` and it adds that annotation. I have also checked and without annotation translation doesn't work, so I'll leave it as it is.

modules/thunder_updater/src/Command/GenerateThunderUpdateCommand.php
mtodor: 👍

PR #461, opened 2017-11-30
Fix upgrade path and add update path testing

alexpott:

- [X] All coding styles are fulfilled. ([How to check for cs issues?](https://github.com/BurdaMagazinOrg/thunder-dev-tools/blob/master/README.md#code-style-guidelines))
- [X] All tests are running locally. ([How to run the test?](https://github.com/BurdaMagazinOrg/thunder-distribution/blob/develop/docs/development.md#how-to-run-the-tests))
- [X] Necessary update hooks are provided.
- [X] User roles have correct access for new introduced permission.
- [X] Every thunder module has a README.md in its root. Follow [this guidelines](https://www.drupal.org/node/2181737), but we don't need every topic.


alexpott:
Note that the test will fail unless 8.4.x HEAD is used because we need https://www.drupal.org/project/drupal/issues/2925550 to be applied.

thunder.install
```diff
@@ -29,7 +30,24 @@ function thunder_update_dependencies() {
     8006 => $installThunderUpdaterHook,
   ];
 
-  return $dependencies;
+  // Ensure that all thunder updates come after all the other update hooks.
+  $list = [];
+  $return = [];
+  $updates = update_get_update_list();
+  $module_data = system_rebuild_module_data();
+  $thunder_path_regex = '@^' . $module_data['thunder']->getPath() . '@';
```
mtodor: One question, when we use `drush make` build then all modules are in profile `modules` directory too. Fe. `.../thunder/profiles/thunder/modules`. Could that be problem?

alexpott: Totally will be a problem - nice spot. I guess a hardcoded list - I thought about matching on thunder_ but this feels icky.

alexpott:
@dbosen I've addressed the db dump in the repository by moving it to AWS and adding a ThunderAwsTestFixtureTrait to handle getting stuff from there.

thunder.install
alexpott: Latest commit has changed this to a hard-coded list.

drupal-org-core.make
```diff
@@ -2,4 +2,3 @@ api = 2
 core = 8.x
 projects[drupal][type] = core
 projects[drupal][version] = 8.4.2
```
dbosen: The Drupal Version has to be updated in drupal-org-core.make

steffenr:
Thanks for the patch - we also run into an issue cause thunder_updated wanted to do things before main drupal updates where run. The updates from thunder_update_dependencies fixed those issues.

PR #466, opened 2017-12-13
Fix/2926797 improve paragraph split test

jakubhnilicka:

- [ x] All coding styles are fulfilled. ([How to check for cs issues?](https://github.com/BurdaMagazinOrg/thunder-dev-tools/blob/master/README.md#code-style-guidelines))
- [ x] All tests are running locally. ([How to run the test?](https://github.com/BurdaMagazinOrg/thunder-distribution/blob/develop/docs/development.md#how-to-run-the-tests))
- [x ] Code is covered with well-balanced amount of inline comments.


build-thunder.make
```diff
@@ -3,4 +3,4 @@ core = 8.x
 includes[] = drupal-org-core.make
 projects[thunder][type] = profile
 projects[thunder][download][type] = git
-projects[thunder][download][branch] = develop
+projects[thunder][download][tag] = 8.2.10
```
dbosen: Please revert this change

mtodor:
I have tested new changes a bit. It looks really nice, all edge cases we had experienced before are good now. So I'm not able to get into some state where data is lost or that it behaves strange.

But I have found one new edge-edge case, when you do double click on "split before", then data is lost. It doesn't happen for "split after", I'm not sure what is reason. It could be because it's possible to "split before" empty content and it's not possible to "split after" empty content.

PR #480, opened 2018-01-17
Issue #2937285: run updates of enabled modules only.

stborchert:
When running `thunder_update_8112()` some other update functions are called. Unfortunately the updater does not check if the module of the requested function is enabled. If (i.e.) you don't have _thunder_article_ enabled, the update will fail.

tldr; Fixes https://www.drupal.org/project/thunder/issues/2937285.

### DIFF
--- a/modules/thunder_updater/src/Updater.php
+++ b/modules/thunder_updater/src/Updater.php
@@ -7,6 +7,7 @@ use Drupal\config_update\ConfigRevertInterface;
 use Drupal\Core\Config\ConfigFactoryInterface;
 use Drupal\Core\Extension\MissingDependencyException;
 use Drupal\Core\Extension\ModuleInstallerInterface;
+use Drupal\Core\Extension\ModuleHandlerInterface;
 use Drupal\Core\StringTranslation\StringTranslationTrait;
 use Drupal\user\SharedTempStoreFactory;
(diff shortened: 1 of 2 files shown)


mtodor:
@stborchert Thank you for reporting problem.

I'm not sure this is correct way to solve this problem. Ignoring some parts of configuration updates could lead to incorrect behaviour of introduced functionality or made changes. Currently we have extensive logging of failed configuration updates and with this change they are silently ignored.

Also we want to have one module where all configuration updates and update hooks will be collected. That means `thunder_update_8112()` hook has wrong approach.

I can't propose solution like this for following problem, it should not ever happen that update hook is executed and required configuration update definition files are not available for it. I would rather move all configuration update files into profile, where update hook is.

stborchert:
Hey.
Ignoring updates is not the goal of this PR (and is not, what it does).
If you run the update without having _thunder_article_ or _thunder_paragraphs_ enabled, the update fails, because the updater tries to call the update functions of these modules.

With the PR the updater only runs updates of enabled modules.

mtodor:
@stborchert I have made other PR for this problem: #525 - can you please take a look and test for your problem.

mtodor:
We have merged #525.

So I'll close this PR.

PR #481, opened 2018-01-18
Fix thunder_update_8107()

stborchert:
### Problem description

`thunder_update_8107()` installs the module _thunder_riddle_ that creates the new entity types "media.riddle" and "paragraphs.riddle". 
After installation the update tries to convert old paragraphs from _paragraphs_riddle_marketplace_ to the new paragraphs type. The update searches for media entities of bundle "riddle" and the specified riddle ID. Unfortunately there are no media entities of this type yet, so creating the new paragraphs (and all other steps of the update) fails.

### Solution

Create new media entities using the proper values if they do not exist.

### DIFF
--- a/thunder.install
+++ b/thunder.install
@@ -459,70 +459,99 @@ function thunder_update_8107() {
           'type' => 'paragraphs_riddle_marketplace',
         ]);
 
+      /** @var \Drupal\paragraphs\ParagraphInterface $oldParagraph */
       foreach ($paragraphs as $oldParagraph) {
-
         $fieldName = $oldParagraph->get('parent_field_name')->get(0)->getString();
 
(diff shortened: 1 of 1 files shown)


chrfritsch:
Why are the media entities not there? The first thing after the module is enabled i calling       \Drupal::service('media_riddle_marketplace')->createMediaEntities();. That imports all the riddles of your account.

stborchert:
We are migrating from _riddle_marketplace-8.x-1.x_ (and _thunder-8.1.5_) which did not have any media bundles (as far as I can tell). The riddle paragraph we are using (and the update is trying to convert) simply has a link field with the URL to a riddle.

stborchert:
Using the latest development version of _thunder_ fixes the problem (whatever the problem has been :) ).

PR #490, opened 2018-02-01
Select specific config to be enabled on module install

alexpott:

Drupal.org issue: https://www.drupal.org/project/thunder/issues/2941312

### DIFF
--- a/config/schema/thunder.schema.yml
+++ b/config/schema/thunder.schema.yml
@@ -1,0 +1,20 @@
+thunder_third_party:
+  type: mapping
+  label: 'Schema for Thunder profile additions to configuration entities'
+  mapping:
+    config_select:
+      type: mapping
+      label: 'Information for Thunder to select which configuration to enable'
+      mapping:
+        feature:
+          type: string
+          label: 'The Thunder feature name'
+        priority:
+          type: integer
+          label: 'The Thunder feature priority'
+
+block.block.*.third_party.thunder:
+  type: thunder_third_party
+
+views.view.*.third_party.thunder:
+  type: thunder_third_party
--- a/src/ConfigSelector.php
+++ b/src/ConfigSelector.php
@@ -90,12 +90,24 @@ class ConfigSelector {
    * The list makes it simple to work out what configuration is new and if we
    * have to enable or disable any configuration.
    *
+   * @param string $module
+   *   The module being installed.
+   *
    * @return $this
    *
    * @see thunder_module_preinstall()
    */
-  public function setCurrentConfigList() {
-    $this->state->set('thunder.current_config_list', $this->configFactory->listAll());
+  public function setCurrentConfigList($module) {
+    if ($module === 'thunder') {
+      // If the Thunder install profile is being installed we need to process
+      // all existing configuration in
+      // \Drupal\thunder\ConfigSelector::selectConfig().
+      $list = [];
+    }
+    else {
+      $list = $this->configFactory->listAll();
+    }
+    $this->state->set('thunder.current_config_list', $list);
     return $this;
   }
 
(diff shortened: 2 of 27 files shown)


chrfritsch:
That's really nice.

I tried the following:

1. Duplicated the content view and removed all the content lock fields in the new view.
2. Then I added 
```
third_party_settings:
  thunder:
    config_select:
      feature: feature_a
      priority: 0
```
to the existing content view and 
```
third_party_settings:
  thunder:
    config_select:
      feature: feature_a
      priority: 5
```
to the new view.
3. A fresh Thunder installation
4. Disabled content_lock via drush
5. Enter /admin/content, I got
`Exception: No entity type for field name on view content in Drupal\views\Plugin\views\HandlerBase->getEntityType() (line 711 of /docroot/core/modules/views/src/Plugin/views/HandlerBase.php).`
6. Enter admin/structure/views/view/content_foo (the new view) and resave the view.
7. Enter /admin/content again and everything works fine.

Note: The content view with content_lock fields is still enabled, maybe we should disable it.

alexpott:
@chrfritsch that looks like a problem with the view missing dependencies - which does not entirely surprise me looking at content_lock_views_data() so adding the dependency either via a fix to content_lock or via an enforced_dependency is going to be the only way to fix what you've reported.

dbosen:
This seems not to work, when the configuration is put into the distributions config folder. It just works, when the configuration is put into a module. Is that intentional?

alexpott:
Nope that is not intentional. In fact it working in profiles is important.
Nice catch - will look at it on Wednesday.

dbosen:
@alexpott I created a branch, that contains the configuration, I tested with. I just modified the views.view.content.yml and added a new view in views.view.content_lock.yml.
If you checkout the branch test/config-selector-fails-in-profile and install with this, you will see both views get installed and both are enabled. I would expect one view to be disabled.

mtodor:
@alexpott I have pushed one additional test. It tests indirect module uninstall.

If you change dependancy in `config_test.dynamic.feature_a_depends_on_test_two.yml` to depend on module `thunder_config_test_two` and then you run only `testConfigSelectorIndirectDependency()` - you will get correct result, but not if dependency is defined on depending module.

dbosen:
@alexpott I have found something else, but I am not sure, if it actually relates to this PR or is related to the configuration I am using.
I put it into an branch as well: 'test/search-api-config-not-used' 
What I did there is to create a new module "thunder_content_list", that contains three different versions of the content list view.

1. With Content lock (moved from profile config directory, prio 5)
2. Without content lock (newly created, as in the other branch, prio 0)
3. With Search API (plus some more config that is necessary, prio 10)

If you install, you will get the default content list view from the node module, because the new module is not active. So before doing anything else, remove the default view:

`drush ev "\Drupal\views\Entity\View::load('content')->delete()"`

Now you have two possibilities, that should have the same result, but don't:

- first enable search_api_db module
- then enable thunder_content_list module 

This installs as expected: all three views are imported and the search API view is the only active one.
But if you install the other way round - thunder_content_list first, then search_api_db - the search api view does not get installed and the content lock view is still the active one.

alexpott:
@dbosen the issue you've tested and created 'test/search-api-config-not-used' is a case of https://www.drupal.org/project/drupal/issues/2930996 - I've pushed a quick fix with enforced dependencies but we need to get the core issue done. It's not related to this at all. As you noted in HEAD the content does not get created at all.

dbosen:
@alexpott ah, that issue. Yes, I was already unsure about this beeing a problem with the config selector.

alexpott:
Closing in favour of https://github.com/BurdaMagazinOrg/thunder-distribution/pull/494

PR #494, opened 2018-02-14
Issue #2941312 by alexpott: Use Configuration Selector module in Thunder

alexpott:


mtodor:
Just writing few things that we could improve on PR before merging:
1. we should release module (alpha1 or so) , so that we can use nicer versions in composer + make
2. entry in update checklist so that it's transparent for user what is going on. And also add in update hook marking of successful update for that checklist entry.
3. [consideration] maybe we should think about naming post update hooks. We could use similar naming as for normal update hooks - fe. thunder version + counter => `_2_1_16_0`, or something like that (we have nicer description of update hook in comment anyway, that's why function names can be generic). In that way, we can easier integrate update helper in future.

alexpott:
Addressed @mtodor's review. There's now an alpha1 and composer.json and the make file have been updated accordingly. I've integrated the post update into updates.yml.

We don't need a counter on post update hooks they are not related to schema or anything. We could for convenience add the targeted thunder version if we want but event this is unnecessary the number would do nothing.

PR #495, opened 2018-02-19
Content lock with config selector integration

chrfritsch:


chrfritsch:
Updating
- Add config selection settings to existing view
- Import the content_lock view
- Newly created views have to be disabled
- Implement hook_modules_uninstall, to import the basic content view if there is none after uninstalling a module

thunder.post_update.php
```diff
@@ -22,3 +24,81 @@ function thunder_post_update_ensure_config_selector_installed() {
...
+      }
+    }
+  }
+
+  if ($view_contains_content_lock) {
+    // Duplicate content into content_content_lock view and add third party
+    // settings.
+    $content_lock_view = $content_view->createDuplicate();
+    $content_lock_view->set('id', 'content_content_lock');
+    $content_lock_view->setThirdPartySetting('config_selector', 'feature', 'thunder_content_view');
```
alexpott: Not sure we need reset the feature. I think it might clearer without this.

thunder.post_update.php
```diff
@@ -22,3 +24,81 @@ function thunder_post_update_ensure_config_selector_installed() {
...
+        if (!empty($display['display_options'][$option]) && is_array($display['display_options'][$option])) {
+          foreach ($display['display_options'][$option] as $key => $values) {
+            if ($values['table'] == 'content_lock' || (!empty($values['relationship']) && in_array($values['relationship'], $deleted_relationships))) {
+              unset($display['display_options'][$option][$key]);
+            }
+          }
+        }
+      }
+    }
+    $content_view->disable();
```
alexpott: Let's add a comment here that we're disabling with view without content lock. Because the content lock view is going to take precedence. Actually don't we want to import the non content lock view here? So they have a view to fallback to if they every uninstall content_lock? Atm this view will still depend on content_lock.

alexpott: Ah I see you are removing all the content_lock stuff above. Any reason why we don't just use the view that comes with Thunder?

thunder.post_update.php
```diff
@@ -22,3 +24,81 @@ function thunder_post_update_ensure_config_selector_installed() {
...
+
+  $view_contains_content_lock = FALSE;
+  foreach (array_keys($content_view->get('display')) as $display_name) {
+    $display = $content_view->getDisplay($display_name);
+    $options = ['fields', 'filters', 'sorts', 'relationships'];
+    foreach ($options as $option) {
+      if (!empty($display['display_options'][$option]) && is_array($display['display_options'][$option])) {
+        foreach ($display['display_options'][$option] as $key => $values) {
+          if ($values['table'] == 'content_lock') {
+            $view_contains_content_lock = TRUE;
```
alexpott: Rather than doing this we could just calculate dependencies and check if content_lock module is there no?

thunder.profile
```diff
@@ -15,7 +15,7 @@ use Drupal\block\Entity\Block;
 function thunder_system_info_alter(array &$info, Extension $file, $type) {
   // Thunder can not work properly without these modules. So they are enforced
   // to be enabled.
-  $required_modules = ['config_selector'];
+  $required_modules = ['config_selector', 'views', 'media_entity', 'node'];
```
alexpott: Like it. Profiles having dependencies ftw!

alexpott (commented):
I think the whole update function can be simplified a bit. It feels like it is on the right lines it just has to decide what the current view is - and then decide which view from thunder install config to import.

thunder.post_update.php
chrfritsch: Exactly, we want that the view without the content_lock fields is very close to what had the user before.

thunder.post_update.php
chrfritsch: Good point 👍

thunder.post_update.php
dbosen: Its not actually resetting, its adding the third party settings to the original view, that did not have it.

alexpott: It does though - we're adding them above.

thunder.post_update.php
```diff
@@ -22,3 +24,70 @@ function thunder_post_update_ensure_config_selector_installed() {
...
+  if (in_array('content_lock', $content_view->getDependencies()['module'])) {
+    // Duplicate content into content_content_lock view and add third party
+    // settings.
+    $content_lock_view = $content_view->createDuplicate();
+    $content_lock_view->set('id', 'content_content_lock');
+    $content_lock_view->setThirdPartySetting('config_selector', 'feature', 'thunder_content_view');
+    $content_lock_view->setThirdPartySetting('config_selector', 'priority', 1);
+    $content_lock_view->save();
+
+    // Delete all content_lock related stuff from the content view.
```
alexpott: I don't get why we do this. Their current view is now the content lock view - why don't we just import the non-content lock view from the install config and be done.

thunder.post_update.php
dbosen: Yeah, I just missed line 37.

thunder.post_update.php
dbosen: We do that to keep the changes a user has done to the view.

thunder.profile
dbosen: Not sure about media_entity though, the dependency will change, once we switch to core media. But that should not be a problem.

chrfritsch:
https://github.com/BurdaMagazinOrg/thunder-distribution/pull/498 will do the thing

PR #498, opened 2018-02-20
Issue #2950317 by chr.fritsch, alexpott: Cleanup content list

chrfritsch:


modules/thunder_article/thunder_article.post_update.php
```diff
@@ -0,0 +1,15 @@
...
+
+/**
+ * @file
+ * Post update functions for Thunder.
+ */
+
+/**
+ * Move scheduler view tab to local task on content list.
+ */
+function thunder_article_post_update_move_scheduler_to_content_list() {
```
alexpott: Should we be using the checklist thing for this update?

thunder.profile
```diff
@@ -334,6 +334,31 @@ function thunder_modules_installed($modules) {
...
 }
 
+/**
+ * Implements hook_modules_uninstalled().
+ */
+function thunder_modules_uninstalled($modules) {
+  // Import the content view if it was deleted during module uninstalling.
+  // This could happen if content_lock was uninstalled and the content view
+  // contained content_lock fields at that time.
+  if (in_array('content_lock', $modules)) {
```
alexpott: Always best to use the third argument of in_array()... just set it to TRUE for type safe checking - otherwise some truly odd things are possible. See, for example, https://www.drupal.org/forum/support/module-development-and-code-questions/2009-09-10/epic-wtf-with-in_array-please-help

modules/thunder_article/thunder_article.links.task.yml
```diff
@@ -0,0 +1,9 @@
+thunder_article.content_view:
+  deriver: 'Drupal\thunder_article\Plugin\Derivative\DynamicLocalTasks'
+  weight: 100
+
+# We need an 'Overview' task also when content_moderation is not enabled
+content_moderation.content:
```
alexpott: Funky.

modules/thunder_article/thunder_article.post_update.php
chrfritsch: I also asked @dbosen about that and we think it's fine to leave it out here. We are just importing a simple config and not modifying an existing one. 

@alexpott Could you think about a case, where this might fail?

alexpott: I thought the point of the checklist was not only to tell people about things that might fail but also to tell people what happened. If it is only about possible fails I agree - no need.

dbosen: Yes, for the previous pull request it was not so important. But now we also replace  the locked_content view. This should be communicated in the checklist thing.
And also I am not as light headed as I was yesterday ;)

modules/thunder_article/src/Form/ConfigurationForm.php
```diff
@@ -0,0 +1,99 @@
...
+   */
+  public function submitForm(array &$form, FormStateInterface $form_state) {
+    parent::submitForm($form, $form_state);
+
+    $this->config('thunder_article.settings')
+      ->set('move_scheduler_local_task', $form_state->getValue('move_scheduler_local_task'))
+      ->save();
+
+    $this->localTaskLinkManager->clearCachedDefinitions();
+    $this->cacheRender->invalidateAll();
```
dbosen: This still seems to be rather harsh. While I do not suppose, that this gets changed very often, it could still invalidate several GB of cached data.

modules/thunder_updater/updates.yml
```diff
@@ -200,3 +200,12 @@ v2.14:
...
         extension_page: { '#text': 'Enable the Configuration Selector module', '#url': 'internal:/admin/modules' }
+v2.16:
+    '#title': 'Update to Thunder 2.16'
+    '#description': '<h3>These are the new features for this release:</h3>'
+    thunder_article_move_scheduler_to_content_list:
+        '#title': 'Update locked content view and move scheduler tab to local tasks'
+        '#description': '<p>From now on we will not add functionality for content lists to the existing content list but will create a separate content list for each use-case.</p><p>We have replaced the current locked content view with a new one. The old one can still be reached under views.view.locked_content_old.</p><p>The scheduler tab has also been moved to the local tasks of the content list. This change can be undone under the menu item "Thunder article configuration".</p>'
+        '#description_successful': '<p>The installation was successful.</p>'
+        '#description_failed': '<p><strong>Installation failed</strong><p>'
+        thunder_article_config: { '#text': 'Revert scheduler into local tasks', '#url': 'internal:/admin/config/thunder_article/configuration' }
```
alexpott: Revert is a funny word. I'd write this as "Move scheduler back to local tasks"

modules/thunder_article/src/Form/ConfigurationForm.php
alexpott: I think there might be a way around this.

modules/thunder_article/thunder_article.module
```diff
@@ -43,16 +43,11 @@ function _thunder_article_form_alter_helper(array &$form, FormStateInterface $fo
 function thunder_article_menu_local_tasks_alter(&$data, $route_name) {
   foreach ($data['tabs'] as $key => $tab) {
     foreach ($tab as $task => $link) {
-      if ($task === 'views_view:view.scheduler_scheduled_content.overview') {
+      $tabs = ['views_view:view.scheduler_scheduled_content.overview', 'thunder_article.content_view:thunder_article.scheduler'];
+      if (in_array($task, $tabs, TRUE)) {
         $config = \Drupal::service('config.factory')->get('thunder_article.settings');
```
dbosen: could we get this line, as well as line 46, in front of the foreaches?

alexpott: So the reason I didn't do that is that that every local_task_alter will do a query. This way we only do 1 query when we need to

PR #499, opened 2018-02-21
Issue #2947051 by chr.fritsch, alexpott: Config Selector and Thunder Updater are required by Thunder

chrfritsch:

### DIFF
--- a/thunder.post_update.php
+++ b/thunder.post_update.php
@@ -6,7 +6,7 @@
  */
 
 /**
- * Install the Configuration Selector module.
+ * Install the Configuration Selector module if necessary.
  */
 function thunder_post_update_ensure_config_selector_installed() {
   /** @var \Drupal\thunder_updater\Updater $thunderUpdater */
--- a/thunder.post_update.php
+++ b/thunder.post_update.php
@@ -12,6 +12,10 @@ function thunder_post_update_ensure_config_selector_installed() {
   /** @var \Drupal\thunder_updater\Updater $thunderUpdater */
   $thunderUpdater = \Drupal::service('thunder_updater');
   if (!\Drupal::moduleHandler()->moduleExists('config_selector')) {
+    // This function was renamed after Thunder 2.14, because we want to ensure
+    // that the config_selector is always enabled.
+    // thunder_post_update_install_config_selector was the former name and is
+    // still the key for the checklist.
     $thunderUpdater->installModules(['thunder_post_update_install_config_selector' => 'config_selector']);
   }
 
--- a/thunder.post_update.php
+++ b/thunder.post_update.php
@@ -8,10 +8,12 @@
 /**
  * Install the Configuration Selector module.
  */
-function thunder_post_update_install_config_selector() {
+function thunder_post_update_ensure_config_selector_installed() {
   /** @var \Drupal\thunder_updater\Updater $thunderUpdater */
   $thunderUpdater = \Drupal::service('thunder_updater');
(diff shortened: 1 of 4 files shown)


alexpott:
I doubt Dries knows... or God for that matter.

thunder.profile
```diff
@@ -5,9 +5,22 @@
...
 
+/**
+ * Implements hook_system_info_alter().
+ */
+function thunder_system_info_alter(array &$info, Extension $file, $type) {
+  // Thunder can not work properly without these modules. So they are enforced
+  // to be enabled.
+  $required_modules = ['config_selector'];
+  if ($type == 'module' && in_array($file->getName(), $required_modules)) {
+    $info['required'] = TRUE;
```
alexpott: Often implementations add an explanation here too. But the default reason actually looks good. On the extend screen it says "Required by: Thunder" which is exactly the case. We might in the future want to provide more explicit reasons but adding provision for that is premature.

thunder.profile
```diff
@@ -5,9 +5,22 @@
...
 use Drupal\Core\Form\FormStateInterface;
 use Drupal\block\Entity\Block;
 
+/**
+ * Implements hook_system_info_alter().
+ */
+function thunder_system_info_alter(array &$info, Extension $file, $type) {
+  // Thunder can not work properly without these modules. So they are enforced
+  // to be enabled.
+  $required_modules = ['config_selector'];
```
alexpott: I've ensure that if the config_selector is not installed for any reason this breaks nothing. Once the module is installed this ensures that it can not be uninstalled. Either by the UI or Drush.

alexpott:
It might be worth renaming thunder_post_update_install_config_selector and making it only do the install if the module is not installed just to ensure that the module is still installed. Since sites might have uninstalled it after the last update because we no longer required it.

alexpott:
Maybe rename it to thunder_post_update_ensure_config_selector_installed()

thunder.post_update.php
```diff
@@ -8,10 +8,12 @@
 /**
  * Install the Configuration Selector module.
```
alexpott: Should add "if necessary" - probably will be installed on many Thunder sites.

thunder.post_update.php
```diff
@@ -8,10 +8,12 @@
...
  */
-function thunder_post_update_install_config_selector() {
+function thunder_post_update_ensure_config_selector_installed() {
   /** @var \Drupal\thunder_updater\Updater $thunderUpdater */
   $thunderUpdater = \Drupal::service('thunder_updater');
-  $thunderUpdater->installModules(['thunder_post_update_install_config_selector' => 'config_selector']);
+  if (!\Drupal::moduleHandler()->moduleExists('config_selector')) {
+    $thunderUpdater->installModules(['thunder_post_update_install_config_selector' => 'config_selector']);
```
alexpott: I think keeping the same name here makes sense - this means there should only be one checkbox about installing this module.

alexpott: Perhaps a comment might be useful for our futureselves to understand why.

alexpott: That said this update is going to be a part of 2.15 so maybe we will need a new entry in docroot/profiles/contrib/thunder/modules/thunder_updater/updates.yml - not sure.

chrfritsch: I think we don't need a new checkbox entry. For most of the sites this update hook will do nothing. They will not uninstall the module immediately

alexpott: Okay so let's add a comment about why this doesn't match.

PR #509, opened 2018-03-02
Issue #2820056 by chr.fritsch: Be able to create and save a draft without changing the published article (workflow integration)

chrfritsch:


dbosen:
Please add the state change from draft to archived

dbosen:
The sidebar on the right does not show the current state, just "Published" and "Unpublished", it should show the current state instead.

chrfritsch:
Things to do:
- Scheduler does not work correctly https://www.drupal.org/project/scheduler/issues/2798689#comment-12527743
- Default content get's not imported as published (might be related to some test failures) https://www.drupal.org/project/default_content/issues/2952963#comment-12533105

alexpott:
I've discussed some of the issues we're facing with paragraphs and the other core committers. Catch in particular had some interesting points:
> my suggestion would be don’t integrate content_moderation with Thunder, use workspaces instead, any issues with workspaces you’ll also get with content_moderation but they’ll be feasible to fix without disgusting hacks.

The problems are that a single entity is not often a single entity. We struggling with paragraphs (https://www.drupal.org/project/paragraphs/issues/2807371) but there is also path aliases (https://www.drupal.org/project/drupal/issues/2856363) and taxonomy terms.

composer.json
```diff
@@ -74,6 +74,13 @@
             },
             "drupal/redirect": {
                 "Provided config does not match the config after an export": "https://www.drupal.org/files/issues/2932248-3.patch"
+            },
+            "drupal/scheduler": {
+                "Scheduler integration with core Content moderation": "https://www.drupal.org/files/issues/2018-05-07/scheduler-content-moderation-integration-2798689-63.patch"
+            },
+            "drupal/core": {
+                "Move content_moderation config in standard profile": "https://www.drupal.org/files/issues/2018-03-13/move-content_moderation-config.patch",
+                "UpdatePathTestBase fails to pick up new modules": "https://www.drupal.org/files/issues/2018-05-24/2975081-8.5.3-2.patch"
```
dbosen: This patch does not seem to be relevant for users. Patches that are only relevant for our tests should go into the apply_patches function in https://github.com/BurdaMagazinOrg/thunder-distribution/blob/8.x-2.x/scripts/travis/04-install-thunder.sh 
This should also take care of the failing drush make tests.

IT-Cru:
BUG: Role content creator should not have permission to schedule publishing of content.

dbosen:
Alignment of the selectbox is broken.

bildschirmfoto 2018-09-04 um 10 35 19

dbosen:
Also, if only one option is available, remove the select box and just display the text - in the case of the screenshot in the other comment, the word "Draft should be written instead of the unusable select box.

dbosen:
bildschirmfoto 2018-09-04 um 11 19 37

Change "Delete revision" to "Delete <state>" e.g. "Delete Draft".
Also change the "Published [Draft]" part to "[Draft] of published Article", or in general:
"[Moderation state] of <state> <content type>"

dbosen (changes requested):
See change requests in comments

dbosen:
bildschirmfoto 2018-09-04 um 12 00 13

Button should not be changed to "Save as" if content type is not moderated.

modules/thunder_article/src/Form/ThunderNodeForm.php
```diff
@@ -0,0 +1,139 @@
...
+use Drupal\Core\Entity\EntityTypeBundleInfoInterface;
+use Drupal\Core\Form\FormStateInterface;
+use Drupal\Core\Session\AccountInterface;
+use Drupal\Core\TempStore\PrivateTempStoreFactory;
+use Drupal\Core\Url;
+use Drupal\node\NodeForm;
+use Symfony\Component\DependencyInjection\ContainerInterface;
+
+/**
+ * Base for handler for taxonomy term edit forms.
```
alexpott: A node form no?

composer.json
```diff
@@ -126,7 +129,7 @@
         "drupal/responsive_preview": "^1.0",
         "drupal/redirect": "^1.0",
         "drupal/riddle_marketplace": "^2.0",
-        "drupal/scheduler": "^1.0",
+        "drupal/scheduler": "1.0",
```
alexpott, line 132: How come this change? Is there a version we're incompatible with?

config/install/user.role.restricted_editor.yml
```diff
@@ -0,0 +1,58 @@
...
+  - 'edit terms in tags'
+  - 'never autoplay videos'
+  - 'renew token'
+  - 'revert article revisions'
+  - 'revert page revisions'
+  - 'update any media'
+  - 'update media'
+  - 'use editorial transition create_new_draft'
+  - 'use editorial transition unpublished_draft'
+  - 'use editorial transition unpublished_published'
```
alexpott: If the idea is that this user can't publish anything then maybe they shouldn't have this permission.

src/Plugin/Field/FieldWidget/ModerationStateWidget.php
```diff
@@ -0,0 +1,67 @@
...
+ *
+ * @FieldWidget(
+ *   id = "thunder_moderation_state_default",
+ *   label = @Translation("Moderation state (Thunder)"),
+ *   field_types = {
+ *     "string"
+ *   }
+ * )
+ */
+class ModerationStateWidget extends CoreModerationStateWidget {
```
alexpott, line 24: I think it might work documenting why we are overriding this widget.

alexpott: It appears to be a simply this widget to be a list of possible future states as thunder provides a different way to know the current state.

thunder.info.yml
```diff
@@ -70,7 +70,7 @@ install:
   - paragraphs_features
   - pathauto
   - redirect
-  - scheduler
```
alexpott: I'd leave scheduler here as a separate line as it is going to end up installed. It feels like normal practice here to not rely on dependency install.

composer.json
```diff
@@ -70,6 +70,9 @@
             },
             "drupal/entity_browser": {
                 "Provide inline entity form FieldWidgetDisplay": "https://www.drupal.org/files/issues/2858438_6.patch"
+            },
+            "drupal/scheduler": {
+                "Scheduler integration with core Content moderation": "https://www.drupal.org/files/issues/2018-07-31/2798689-98.patch"
```
alexpott: This is not the latest version of the patch you've made. The latest version seems to have considerable changes. Also I'm concerned about the implementation of the unpublished state in this patch because at the moment it lists both draft and unpublished in the thunder use-case. Draft is not going to un-publish anything. It's going to create a new forward revision exactly the same as the published revision and the content will still be published on the site. I think that that is quite dangerous behaviour.

modules/thunder_article/src/Form/ThunderNodeForm.php
```diff
@@ -0,0 +1,139 @@
...
+        'node_revision' => $entity->getLoadedRevisionId(),
+      ]);
+      if ($this->getRequest()->query->has('destination')) {
+        $query = $route_info->getOption('query');
+        $query['destination'] = $this->getRequest()->query->get('destination');
+        $route_info->setOption('query', $query);
+      }
+      $element['delete_revision'] = [
+        '#type' => 'link',
+        '#title' => $this->t('Delete this version'),
```
alexpott: I'm guessing that main purpose here is to allow users to start from the published revision again. Have we considered offering the ability to create a new draft based on the published revision?

modules/thunder_article/src/Form/ThunderNodeForm.php
```diff
@@ -0,0 +1,139 @@
...
+   * {@inheritdoc}
+   */
+  public function form(array $form, FormStateInterface $form_state) {
+    /** @var \Drupal\Core\Entity\ContentEntityFormInterface $form_object */
+    $form_object = $form_state->getFormObject();
+    /** @var \Drupal\node\NodeInterface $entity */
+    $entity = $form_object->getEntity();
+
+    if ($this->moderationInfo->hasPendingRevision($entity)) {
+      $user = $this->entityTypeManager->getStorage('user')->load($entity->getRevisionUserId());
```
alexpott: What's wrong with $entity->getRevisionUser()?

alexpott (commented):
I've manually tested the functionality and it looks pretty good in terms of being quite clear for the user. Which is probably the trickiest part of this.

There is a gotcha with content moderation that we might want to make clear. Somehow. Path aliases are not versioned. So if you change the path alias on a draft and there is published content - the path alias changes for that too :(

composer.json
dbosen: We have the policy, that patched modules have to be version restricted.

chrfritsch: We have to pin the scheduler version until our patch gets committed and released. If we are not going to do this we get composer errors as soon as the patch landed.

modules/thunder_article/src/Form/ThunderNodeForm.php
dbosen: The purpose here is to be able to quickly dismiss the current draft, not necessarily to create a new one. But the use case of creating a new draft (that overrides an existing one) might be interesting as well, I will discuss this when we present the current state to our stake holders.

alexpott: But doesn't this get into misleading territory when there are multiple forward revisions? In a revisioning world the ideal thing is to not to do deletes.

modules/thunder_article/src/Form/ThunderNodeForm.php
chrfritsch: Didn't know that.

thunder.info.yml
chrfritsch: Fixed

src/Plugin/Field/FieldWidget/ModerationStateWidget.php
chrfritsch: Fixed

config/install/user.role.restricted_editor.yml
chrfritsch: True story. Removed the 'use editorial transition unpublished_published' permission

modules/thunder_article/src/Form/ThunderNodeForm.php
chrfritsch: Fixed

modules/thunder_article/src/Form/ThunderNodeForm.php
dbosen: Yes, that is a problem.

dbosen: The current idea is to change this to something like "Revert to last published version"

alexpott: 👍 That sounds like a good idea because:
* Non destructive
* Less deletes all in a row in the UI.

modules/thunder_article/src/Form/ThunderNodeForm.php
```diff
@@ -0,0 +1,138 @@
...
+   * {@inheritdoc}
+   */
+  public function form(array $form, FormStateInterface $form_state) {
+    /** @var \Drupal\Core\Entity\ContentEntityFormInterface $form_object */
+    $form_object = $form_state->getFormObject();
+    /** @var \Drupal\node\NodeInterface $entity */
+    $entity = $form_object->getEntity();
+
+    if ($this->moderationInfo->hasPendingRevision($entity)) {
+      $this->messenger()->addWarning($this->t('This %entity_type has unpublished changes from user %user.', ['%entity_type' => $entity->get('type')->entity->label(), '%user' => $entity->getRevisionUser()->label()]));
```
alexpott, line 73: Just thought about this some more. I think we should also copy better with multiple forward revisions by directing people to the list of revisions. This will also allow people to compare the revisions to see has has changed.

klucic:
Hi All we just had a Review with Steffen and Mary and got the following Feedback

- ~~emphasize more that a Node is in "Draft Mode", as the current implementation is functionally working but is easy to oversee~~
- [x] deleting the current version should be replaced by the option to revert to the live version
- [x] remove redundand information of revision and publishing state (published of published, unpublished of unpublished)

src/Form/NodeRevisionRevertDefaultForm.php
```diff
@@ -0,0 +1,90 @@
...
+use Drupal\Core\Url;
+use Drupal\node\Form\NodeRevisionRevertForm;
+use Drupal\node\NodeInterface;
+
+/**
+ * Provides a form for reverting a node revision.
+ *
+ * @internal
+ */
+class NodeRevisionRevertDefaultForm extends NodeRevisionRevertForm {
```
alexpott, line 16: I think this functionality works really nicely.

alexpott: We do seem to be missing automated test coverage of this functionality though.

alexpott (commented):
Manual testing shows everything works well. I do find the whole mixing of creating drafts and publishing and unpublishing odd for simple workflows like this. But that's part of having an extensible system.

dbosen (approved):

PR #538, opened 2018-04-20
Issue #2969459 by chr.fritsch: Remove custom code for redirect after installation

chrfritsch:


alexpott:
We need 8.6.x right? So this has to wait.

chrfritsch:
This is the Thunder 3.x branch. We are using 8.6.x branch here because it will not be finished before that is released ;)

dbosen:
But then we should depend on 8.6.x in the info.yml (is that possible?)

chrfritsch:
Looks like drupal_verify_profile() is not able to handle something like   
`- system (>=8.6.0-dev)`
in the profile.info.yml

chrfritsch:
We have the 8.6.x dependency in composer.json and in drupal-org-core.make. So I think we are fine.

mtodor (approved):

PR #543, opened 2018-04-26
Issue #2969454 by chr.fritsch: Use core functionality to define real profile dependencies

chrfritsch:

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -71,6 +71,9 @@
             },
             "drupal/redirect": {
                 "Provided config does not match the config after an export": "https://www.drupal.org/files/issues/2932248-3.patch"
+            },
+            "drupal/core": {
+                "Allow install profiles to require modules": "https://www.drupal.org/files/issues/2018-04-09/2952888-25.patch"
             }
         }
     },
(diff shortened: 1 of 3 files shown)


thunder.info.yml
```diff
@@ -9,6 +9,12 @@ distribution:
 
 # Required modules
 dependencies:
+  - config_selector
+  - media
+  - node
+  - views
```
dbosen, line 15: I am unsure about the views dependency. It is possible to do meaningful sites without views and we just need them for the content list, which would fall back to the entity list without views.

composer.json
```diff
@@ -71,6 +71,9 @@
             },
             "drupal/redirect": {
                 "Provided config does not match the config after an export": "https://www.drupal.org/files/issues/2932248-3.patch"
+            },
+            "drupal/core": {
+                "Allow install profiles to require modules": "https://www.drupal.org/files/issues/2018-04-09/2952888-25.patch"
```
dbosen: This has to be added to the drush make file as well

thunder.info.yml
chrfritsch: This PR is for replacing the existing behavior with core functionality. For changing the behavior we should create a new PR and define which dependencies are useful

mtodor (approved):

PR #559, opened 2018-06-12
Issue #2979303 by chr.fritsch: Use exclude option for paragraphs allowed types

chrfritsch:


thunder.profile
```diff
@@ -319,22 +319,6 @@ function thunder_modules_installed($modules) {
...
-    $settings = $field->getSetting('handler_settings');
-
-    $settings['target_bundles']['riddle'] = 'riddle';
-    $settings['target_bundles_drag_drop']['riddle'] = ['enabled' => TRUE, 'weight' => 10];
-
-    $field->setSetting('handler_settings', $settings);
-
-    $field->save();
-  }
-
```
dbosen: we should add disabling this option just in the integration test, so it does not fail

dbosen (changes requested):

dbosen (approved):

PR #598, opened 2018-10-09
Issue #3005411: Default users for testing

mtodor:


modules/thunder_demo/content/user/09ec113d-32a1-4dcf-80d9-fc78069ce7b1.json
```diff
@@ -0,0 +1,101 @@
...
+        }
+    ],
+    "preferred_admin_langcode": [
+        {
+            "value": "en"
+        }
+    ],
+    "name": [
+        {
+            "value": "kayne"
```
dbosen: I guess this should be Kanye :) - Well, actually it is just Ye now (https://en.wikipedia.org/wiki/Kanye_West#Name_%22Ye%22).

modules/thunder_demo/content/user/09ec113d-32a1-4dcf-80d9-fc78069ce7b1.json
```diff
@@ -0,0 +1,101 @@
...
+        {
+            "value": true,
+            "lang": "en"
+        }
+    ],
+    "metatag": [
+        {
+            "value": {
+                "title": "kayne | Site-Install",
+                "canonical_url": "http:\/\/default\/user\/4",
```
dbosen: This is usually wrong, can we skip the metatag information? The node exports just contain the field_meta_tags values.

modules/thunder_demo/content/user/09ec113d-32a1-4dcf-80d9-fc78069ce7b1.json
```diff
@@ -0,0 +1,101 @@
...
+    ],
+    "roles": [
+        {
+            "target_id": "administrator"
+        }
+    ],
+    "default_langcode": [
+        {
+            "value": true,
+            "lang": "en"
```
dbosen: Lang codes have to be removed everywhere, see node export.

dbosen (changes requested):
I am not sure about this. I did not know, that so many user should be generated, with all the funny names :).
In the distribution we should just have one user per role with simple technical names. Editor, restricted-editor, seo. 
The more complex setups like this should be in the testing only module.

modules/thunder_demo/content/user/09ec113d-32a1-4dcf-80d9-fc78069ce7b1.json
mtodor: Ticket is defined on 24.09.2018. It as legacy naming. 🤣

mtodor:
@dbosen I have addressed all the comments from the review. I've started work on a custom module with all listed testing users (btw. we need repo for that). And here we have one user per provided role. I would keep prefix `test-` before role to avoid possible conflicts with created `admin` or `administrator` user during installation process.

chrfritsch:
Could we open the context of this PR a bit and also add some tags, that are connected to our articles?

That would be helpful for normal testing and also to see some differences on https://github.com/BurdaMagazinOrg/theme-thunder-admin/pull/166

mtodor:
@chrfritsch - I would create new issue for adding of Tags for articles and make it in another PR. It's better to keep scopes smaller.

chrfritsch (dismissed):
Please, then make it so.

We also need a follow-up to set the new users as authors of our nodes.

mtodor:
Added issue for tags: https://www.drupal.org/project/thunder/issues/3008589
And for users assignment: https://www.drupal.org/project/thunder/issues/3008594

dbosen (approved):

PR #599, opened 2018-10-10
Issue #2986591 by daniel.bosen: Implement password policy

dbosen:


config/optional/password_policy.settings.yml
```diff
@@ -0,0 +1 @@
+cron_threshold: 250
```
chrfritsch: This is the same config like provided from the module. There is no need for us to ship this file

thunder.profile
```diff
@@ -271,6 +271,7 @@ function thunder_modules_installed($modules) {
       \Drupal::service('module_installer')->install(['scheduler_content_moderation_integration']);
     }
   }
+
```
chrfritsch: Not needed

thunder.profile
```diff
@@ -321,6 +322,14 @@ function thunder_modules_installed($modules) {
 
     $field->save();
   }
+
+  // When enabling password policy, enabled required sub modules.
+  if (in_array('password_policy', $modules)) {
+    \Drupal::service('module_installer')->install(['password_policy_length']);
+    \Drupal::service('module_installer')->install(['password_policy_history']);
+    \Drupal::service('module_installer')->install(['password_policy_character_types']);
+  }
```
chrfritsch, line 331: Maybe we should write a message that we enabled some additional submodules.

config/optional/field.field.user.user.field_password_expiration.yml
```diff
@@ -0,0 +1,26 @@
...
+field_name: field_password_expiration
+entity_type: user
+bundle: user
+label: 'Password Expiration'
+description: 'The password will expire based on the number of days defined in the policy'
+required: false
+translatable: false
+default_value:
+  -
+    value: 0
```
chrfritsch: Shouldn't this be checked by default?

composer.json
```diff
@@ -125,6 +128,7 @@
         "drupal/metatag": "^1.0",
         "drupal/paragraphs": "1.3",
         "drupal/paragraphs_features": "^1.0",
+        "drupal/password_policy": "^3.0",
```
chrfritsch: We are using a patch. So this should be a fixed version

chrfritsch: Also the module is not added to the makefile. Or should we just add this module to the composer suggestion?

config/optional/core.entity_form_display.user.user.default.yml
```diff
@@ -0,0 +1,53 @@
+langcode: en
```
chrfritsch: Why do we need this file now?

tests/src/Functional/Integration/PasswordPolicyTest.php
```diff
@@ -0,0 +1,74 @@
...
+    'password_policy',
+  ];
+
+  /**
+   * Tests redirect from old URL to new one.
+   */
+  public function testPasswordPolicies() {
+    $editor = $this->drupalCreateUser();
+    $editor->addRole('editor');
+    $editor->save();
```
chrfritsch: You can use $this->logWithRole() instead

chrfritsch (changes requested):

composer.json
dbosen: Good idea, but as you said, when using a patch the version should be fixed, with suggestions, we cannot do that.
I will add it to the makefile as well for now.

config/optional/core.entity_form_display.user.user.default.yml
dbosen: Because of the password fields, but I will try to install without providing this explicitely

tests/src/Functional/Integration/PasswordPolicyTest.php
dbosen: nice!

dbosen: No I can't I need the $editor, which I do not get from logWithRole()

chrfritsch: Then just let logWithRole() return the created user

chrfritsch (dismissed):

mtodor:
image

mtodor (changes requested):
When I run Cron - all passwords are marked as expired. 😞 

How to test it:
- log in as admin when pass policy is enabled, execute cron and you should get Change Pass page

mtodor:
I have noticed one more thing, maybe it's not so relevant, but maybe it can be fixed easy.

How to reproduce it:
1. make pass expired for user with `Editor`, so that you get Change pass form on first login
2. type everything correctly, old pass, new pass, confirm new pass
3. then just add to confirm pass one letter (fe. `a`)
4. after that switch to new pass field and also add that letter, don't click anywhere else, focus has to stay in that field. You will notice that passwords are matching: `Passwords match: yes`.
5. Scroll down and click `Save`
=> you should get error: `The specified passwords do not match.`

For some reason, first password is not sent on form submit.

config/optional/field.field.user.user.field_password_expiration.yml
```diff
@@ -0,0 +1,26 @@
...
+field_name: field_password_expiration
+entity_type: user
+bundle: user
+label: 'Password Expiration'
+description: 'The password will expire based on the number of days defined in the policy'
+required: false
+translatable: false
+default_value:
+  -
+    value: 1
```
mtodor: This option doesn't work when you create new user.

For example:
1. Create new user
1.a. check `Password Expiration` option
1.b. you can also set date in past
2. Save user
3. Edit user
3.a. Option will be unchecked
3.b. Date will be current date

And label is also a bit misleading, because when you edit user and check that it will be automatically marked as expired and it's not related to expire date.

I don't know how to continue on this PR, I would simply switch back to -> Unchecked by default and adjust label to be more clear what will happen when you check it and finish with this "feature". :clap:

mtodor (changes requested):

config/optional/field.field.user.user.field_password_expiration.yml
dbosen: For now, we remove the config and ship with the default where this is not checked. I will try to push the drupal.org issue regarding the misleading wording. See https://www.drupal.org/project/password_policy/issues/2990528

mtodor (dismissed):

chrfritsch (approved):

PR #601, opened 2018-10-10
3005702 Added optional config to place help block on Thunder admin theme

marcelovani:


config/optional/block.block.thunder_admin_help.yml
```diff
@@ -0,0 +1,15 @@
+langcode: en
+status: true
+dependencies: { }
```
dbosen, line 3: We need a dependency on the help module here. Otherwise this gets installed even without the help module enabled, which leads to "This block is broken or missing. You may be missing content or you might need to enable the original module. " Error messages.

dbosen (changes requested):

chrfritsch:
Closing this in favor of https://github.com/BurdaMagazinOrg/theme-thunder-admin/pull/202

PR #617, opened 2018-11-29
Issue #3019383 by chr.fritsch: Remove paragraphs_split_text module

chrfritsch:


dbosen (changes requested):
The update document has to include information, that this functionality is provided by the paragraphs_features module. It should additionally add some information on how to manually update to the paragraphs_features version when the updater failed

chrfritsch:
The documentation is not needed. We already use the functionality provided by paragraphs_features for a while and we also provided an upgrade path.

dbosen (approved):

PR #631, opened 2018-12-14
Issue #3016921: Preview of transition to unpublished state creates er…

chrfritsch:
…rors on edit form


src/Plugin/Field/FieldWidget/ModerationStateWidget.php
```diff
@@ -28,16 +28,27 @@ class ModerationStateWidget extends CoreModerationStateWidget {
    */
   public function formElement(FieldItemListInterface $items, $delta, array $element, array &$form, FormStateInterface $form_state) {
     /** @var \Drupal\Core\Entity\ContentEntityInterface $entity */
-    $entity = $items->getEntity();
+    $entity = $original_entity = $items->getEntity();
```
dbosen, line 31: I think "original_entity" is a bit misleading. The name suggests, that it keeps being unchanged. But instead, it is actually the one that gets changed

dbosen (changes requested):

src/Plugin/Field/FieldWidget/ModerationStateWidget.php
chrfritsch: I used $original_entity, because that's the name that is used in the base class as well.

dbosen: Ok

drupal-org-core.make
```diff
@@ -2,3 +2,4 @@ api = 2
 core = 8.x
 projects[drupal][type] = core
 projects[drupal][version] = 8.6.7
```
dbosen: Version needs to be updated

dbosen (changes requested):

composer.json
```diff
@@ -77,6 +77,9 @@
             "drupal/password_policy": {
                 "Policies should declare a config dependency on the plugin modules": "https://www.drupal.org/files/issues/2018-10-09/2918974-2.patch",
                 "Importing configuration on site without password_policy activated": "https://www.drupal.org/files/issues/2018-07-30/password_policy-config_import_field_error-2771129-57.patch"
+            },
+            "drupal/core": {
+                "Workflow transition current state is changed when using preview mode": "https://www.drupal.org/files/issues/2019-01-28/2914873-32.patch"
```
dbosen, line 82: We have to postpone this until this patch is not necessary anymore

dbosen (changes requested):

PR #637, opened 2019-01-06
Issue #2933086 by truls1502: Update fb_instant_articles from 8.x-1.x to 8.x-2.x

truls1502:


chrfritsch:
Thank you for your contribution!

I think we can't update fb_instant_articles  in Thunder 2, because the way it works is completely different to version 1.x of fb_instant_articles.

So first please make this PR against Thunder 3.x and then it has to be investigated how fb_instant_articles works and how our display modes have to be adopted.

chrfritsch:
Closing in favor of https://github.com/BurdaMagazinOrg/thunder-distribution/pull/680

PR #638, opened 2019-01-07
Issue #3044358: Update travis to PHP 7.3

chrfritsch:


chrfritsch:
Mystery solved:

```
handleUpload uses coninue in a switch: https://cgit.drupalcode.org/dropzonejs/tree/src/UploadHandler.php
Triggering a PHP error:

PHP 7.3 Bug: "continue" targeting switch is equivalent to "break". Did you mean to use "continue 2"

```

alexpott (dismissed):
Looks good. I wonder when we're going to drop PHP 5 support. There is update support for PHP5 in 8.7.x but you can't install on PHP 5 (for what it is worth).

docs/development.md
```diff
@@ -116,7 +116,7 @@ Documentation how to check your code for coding style issues can be found [here]
 
 ## Thunder Travis CI
 
-All Thunder pull requests are execute on [Travis CI](https://travis-ci.org/BurdaMagazinOrg/thunder-distribution). On every pull request tests will be executed (or when new commits are pushed into pull request branch). Tests are executed against PHP versions 5.6 (with drush make install) and 7.2 (with composer install). All code will be checked against coding style.
+All Thunder pull requests are execute on [Travis CI](https://travis-ci.org/BurdaMagazinOrg/thunder-distribution). On every pull request tests will be executed (or when new commits are pushed into pull request branch). Tests are executed against PHP versions 5.6 (with drush make install) and 7.3 (with composer install). All code will be checked against coding style.
```
dbosen: This comment is not correct anymore, we do not test against 5.6 for a while now.

dbosen (changes requested):

dbosen:
> 
> 
> Looks good. I wonder when we're going to drop PHP 5 support. There is update support for PHP5 in 8.7.x but you can't install on PHP 5 (for what it is worth

dbosen (approved):

PR #662, opened 2019-02-26
Issue #3039151 by chr.fritsch, mtodor: Switch to travis stages

chrfritsch:


mtodor:
@chrfritsch can you please create drupal .org issue for this. With problem/motivation, because I would like to understand better what is the reason behind switching to stages. And also that we have one place to discuss the pros/cons and different options.

In general, there are some things that are nice (fe. cleaner separation of the testing environment), but there are others that are not so nice (fe. more jobs running in parallel, executing jobs full time when coding styles are not correct).

chrfritsch:
@mtodor Here is the issue https://www.drupal.org/project/thunder/issues/3039151

.travis.yml
```diff
@@ -29,17 +24,64 @@ addons:
...
+    - stage: sanity checks
+      name: Verify makefile, check coding standards, check for deprecated code
+      env: INSTALL_METHOD=composer
+      install:
+        # Install Drush
+        - composer global require drush/drush:^8.1
+        # Install drupalorg_drush module and verify, that makefile is accepted by drupal.org, otherwise we do not need to go any further
+        - drush dl drupalorg_drush-7.x && drush verify-makefile
+        - bash -x -e ./scripts/travis/03-download-thunder.sh
+      before_script: true
```
mtodor: Let's use `before_script: skip` here as it's in [the documentation](https://docs.travis-ci.com/user/build-stages/#build-stages-and-deployments) because it's more understandable what's happening.

scripts/travis/03-download-thunder-composer.sh
```diff
@@ -0,0 +1,5 @@
+#!/usr/bin/env bash
+
+source ${THUNDER_DIST_DIR}/scripts/travis/functions.sh
```
mtodor: I have looked at `functions.sh` and I do not see the reason for introducing it. I didn't see any function used in two different files. And in my opinion, it introduces additional complexity in code structure. At this point, I don't see a need for shared functions.

.travis.yml
```diff
@@ -29,17 +24,64 @@ addons:
...
+      script:
+        - bash -x -e ./scripts/travis/06-test-source-code.sh
+
+    - stage: sanity checks
+      name: Check config
+      env: INSTALL_METHOD=composer
+      script:
+        - bash -x -e ./scripts/travis/06-run-config-tests.sh
```
mtodor, line 51: This should run for `composer` and `drush_make`.

.travis.yml
```diff
@@ -29,17 +24,64 @@ addons:
...
 
+stages:
+  - sanity checks
+  - test
+
+jobs:
+  include:
+    - stage: sanity checks
+      name: Verify makefile, check coding standards, check for deprecated code
+      env: INSTALL_METHOD=composer
```
mtodor: Can we declare a default for `INSTALL_METHOD` to be `composer`? And then when we need it, we can overwrite it with `env`. I that case we do not have to repeat it everywhere.

I guess some work in `01-setup-environment.sh` would do it.

.travis.yml
```diff
@@ -29,17 +24,64 @@ addons:
...
+      name: Run tests with drush_make build
+      env: INSTALL_METHOD=drush_make
+      install:
+        # Install Drush
+        - composer global require drush/drush:^8.1
+        - bash -x -e ./scripts/travis/03-download-thunder.sh
+
+    - stage: test
+      name: Run update test
+      env: INSTALL_METHOD=composer TEST_UPDATE=true TEST_DEPLOYMENT=true
```
mtodor: We have lost testing of the update for `drush_make`?

`TEST_DEPLOYMENT` was always true, we can just kill it. 😁

scripts/travis/functions.sh
```diff
@@ -0,0 +1,100 @@
...
+    mv ${TEST_DIR}/docroot/sites ${TEST_DIR}/docroot/_sites
+    ln -s ${UPDATE_BASE_PATH}/docroot/sites ${TEST_DIR}/docroot/sites
+
+    cd ${TEST_DIR}/docroot
+
+    # Execute all required updates
+    drush cr
+    drush updatedb -y
+
+    if [[ "${TEST_DEPLOYMENT}" == "true" ]]; then
```
mtodor: Already explained at `.travis.yml` file.

.travis.yml
```diff
@@ -29,17 +24,64 @@ addons:
...
+    - stage: test
+      name: Run tests with composer build
+      env: INSTALL_METHOD=composer
+
+    - stage: test
+      name: Run tests with drush_make build
+      env: INSTALL_METHOD=drush_make
+      install:
+        # Install Drush
+        - composer global require drush/drush:^8.1
```
mtodor: I would pull `composer global require drush/drush:^8.1` into `03-download-thunder.sh`. In that case, we can get rid of `install` step overwrites for 2 stage definitions.

.travis.yml
```diff
@@ -29,17 +24,64 @@ addons:
...
+      env: INSTALL_METHOD=composer
+      install:
+        # Install Drush
+        - composer global require drush/drush:^8.1
+        # Install drupalorg_drush module and verify, that makefile is accepted by drupal.org, otherwise we do not need to go any further
+        - drush dl drupalorg_drush-7.x && drush verify-makefile
+        - bash -x -e ./scripts/travis/03-download-thunder.sh
+      before_script: true
+      script:
+        - bash -x -e ./scripts/travis/06-test-source-code.sh
```
mtodor, line 39: I have to admit, that script naming is a bit confusing for me now. Before we had 00, 01, ... because that was the order how we executed them. But now with 3x`03` and 3x`06`, I'm not sure what is going on.

scripts/travis/03-download-thunder.sh
```diff
@@ -0,0 +1,16 @@
...
+
+# Download latest Thunder release for update
+if [[ ${TEST_UPDATE} == "true" ]]; then
+    # Download latest release from drupal.org
+    drush_download_thunder $UPDATE_BASE_PATH
+fi
+
+# Download Thunder
+if [[ ${INSTALL_METHOD} == "composer" ]]; then
+  source ${THUNDER_DIST_DIR}/scripts/travis/03-download-thunder-composer.sh
```
mtodor: I'm thinking about the option to use these scripts directly in `.travis.yml` file. In that case, we can get rid of `INSTALL_METHOD`. I think that would be cleaner and more readable.

mtodor (changes requested):

.travis.yml
chrfritsch: Yes, we have. Should we bring that back?

mtodor: I don't know. I would like to have Update tests for `drush_make` at least on cron runs. That would lead to 3 => 4 jobs on cron. 😞 
Maybe we can think about excluding some jobs on cron.

scripts/travis/03-download-thunder-composer.sh
```diff
@@ -0,0 +1,22 @@
+#!/usr/bin/env bash
```
mtodor: Can we rename `download-thunder-*` scrips into `build-thunder-*`? Because we are actually building the project there.

mtodor (approved):
Please adjust comment when you do squash and merge - to get rid of all `foo`s, `dsd`s and `dd`s. 😉

PR #668, opened 2019-03-15
Issue #2846679 by chr.fritsch, mtodor: Two authors, but only one shown

chrfritsch:


config/update/thunder_update_8134.yml
```diff
@@ -0,0 +1,218 @@
...
+        uid:
+          settings:
+            width: 100%
+          type: select2_entity_reference
+    delete:
+      content:
+        uid:
+          settings:
+            placeholder: ''
+            size: 60
```
mtodor, line 208: Should we also change the widget for `core.entity_form_display.node.liveblog.default.yml` and `core.entity_form_display.media.riddle.default.yml`?

But if we add that to configuration update file, then the update will fail a lot. Maybe just change the configuration for these form without an update?

config/update/thunder_update_8134.yml
```diff
@@ -0,0 +1,218 @@
...
+    content:
+      uid:
+        settings:
+          placeholder: ''
+          size: 60
+        type: entity_reference_autocomplete
+  update_actions:
+    add:
+      content:
+        field_paragraphs:
```
mtodor: I would exclude these additional fields (`field_paragraphs`, `field_tags`, `field_teaser_media`) from this update.
Problem is if someone has already configured something we will overwrite that. And update has stated something different - `Use select2 widget for author field.`.

I would rather make new updates dedicated to different configuration changes. If we take paragraphs `closed_mode_threshold` as an example. We have that configuration change for taxonomy terms too. 😞

mtodor (changes requested):
Please execute update tests and clean install tests.

mtodor (approved):
Looks good to me

PR #670, opened 2019-03-19
Issue #3041383: Thunder modules installed hook should not expect module installation order

mtodor:

### DIFF
--- a/thunder.profile
+++ b/thunder.profile
@@ -240,33 +240,20 @@ function thunder_themes_installed($theme_list) {
 }
 
 /**
- * Check if provided triggering modules are one of just installed modules.
+ * Check if provided triggering modules are one of the newly installed modules.
  *
- * This function is helper for thunder_modules_installed.
+ * This function is helper for thunder_modules_installed(). Using it in another
+ * context is not recommended. @see hook_modules_installed()
  *
  * @param array $modules
  *   The list of the modules that were newly installed.
  * @param array $triggering_modules
  *   The list of triggering modules required for executing some action.
- * @param bool $only_module_enable
- *   Support only module installations done by module enabling. This flag will
- *   exclude checks for module installation when profile is installed or when
- *   module installation is triggered by configuration import.
  *
  * @return bool
  *   Returns if triggering module is newly installed.
  */
-function _thunder_check_triggering_modules(array $modules, array $triggering_modules, $only_module_enable = TRUE) {
-  // Sometimes we want to react only on module installations done by enabling of
-  // module. In that case we want to ignore module installation when profile is
-  // installed or when module is enabled over configuration import.
-  if (
-    $only_module_enable
-    && (drupal_installation_attempted() || Drupal::isConfigSyncing())
-  ) {
-    return FALSE;
-  }
-
+function _thunder_check_triggering_modules(array $modules, array $triggering_modules) {
   // Check that at least one triggering module is in list of the modules that
   // were newly installed.
   $triggering_not_installed_modules = array_diff($triggering_modules, $modules);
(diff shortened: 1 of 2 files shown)


thunder.profile
```diff
@@ -239,12 +239,54 @@ function thunder_themes_installed($theme_list) {
...
 
+/**
+ * Helper function to check if all required modules are available on install.
+ *
+ * @param array $installed_modules
+ *   Modules installed.
+ * @param array $required_modules
+ *   Required modules to be available.
+ * @param bool $only_direct_module_installs
+ *   The flag to only check if direct module install is executed.
```
alexpott: I don't know what a "direct module install" is

thunder.profile
```diff
@@ -239,12 +239,54 @@ function thunder_themes_installed($theme_list) {
...
+ *
+ * @return bool
+ *   Returns if on module install requirements are fulfilled.
+ */
+function _thunder_are_modules_installed(array $installed_modules, array $required_modules, $only_direct_module_installs = TRUE) {
+  // Check direct module installs.
```
alexpott: This explanation needs expanding and to include why we care about syncing-ness.

thunder.profile
```diff
@@ -239,12 +239,54 @@ function thunder_themes_installed($theme_list) {
   }
 }
 
+/**
+ * Helper function to check if all required modules are available on install.
```
alexpott: are available means are installed right?

thunder.profile
```diff
@@ -239,12 +239,54 @@ function thunder_themes_installed($theme_list) {
...
+ *   Modules installed.
+ * @param array $required_modules
+ *   Required modules to be available.
+ * @param bool $only_direct_module_installs
+ *   The flag to only check if direct module install is executed.
+ *
+ * @return bool
+ *   Returns if on module install requirements are fulfilled.
+ */
+function _thunder_are_modules_installed(array $installed_modules, array $required_modules, $only_direct_module_installs = TRUE) {
```
alexpott: I think we can have a better name that hints at the purpose a bit better - something like _thunder_check_required_modules()

thunder.profile
```diff
@@ -239,12 +239,54 @@ function thunder_themes_installed($theme_list) {
...
+  // Check direct module installs.
+  if (
+    $only_direct_module_installs
+    && (drupal_installation_attempted() || Drupal::isConfigSyncing())
+  ) {
+    return FALSE;
+  }
+
+  // Check that at least one required module is in list of installed modules.
+  $required_not_installed_modules = array_diff($required_modules, $installed_modules);
```
ol0lll: I think $installed_modules name is confusing, because it passes a list of modules currently installing, not the ones installed in the system, which are called active_modules below. 
hook_modules_installed() says
> An array of the modules that were installed.

ol0lll (changes requested):

thunder.profile
```diff
@@ -239,12 +239,54 @@ function thunder_themes_installed($theme_list) {
...
+  if (empty($required_not_installed_modules)) {
+    return TRUE;
+  }
+
+  /** @var \Drupal\Core\Extension\ModuleHandlerInterface $module_handler */
+  $module_handler = Drupal::moduleHandler();
+  $active_modules = array_keys($module_handler->getModuleList());
+
+  // Ensure that all required modules are available.
+  $required_not_active_modules = array_diff($required_not_installed_modules, $active_modules);
```
alexpott: Why is the installed list passed in incorrect?

alexpott: Or to write this a bit different - why are we passing in $modules and not determining the list from the module handler? For a little bit less logic?

alexpott (changes requested):

thunder.profile
mtodor: Ye, it's tricky. It's not the list of installed modules, it's the list of "just installed" modules. That means it's a list of modules that are passed to `hook_modules_installed`.

thunder.profile
```diff
@@ -239,12 +239,62 @@ function thunder_themes_installed($theme_list) {
...
+  $triggering_not_installed_modules = array_diff($triggering_modules, $modules);
+  if (count($triggering_not_installed_modules) === count($triggering_modules)) {
+    return FALSE;
+  }
+
+  // All required triggering modules are in the list of the modules that were
+  // newly installed.
+  if (empty($triggering_not_installed_modules)) {
+    return TRUE;
+  }
```
alexpott, line 268: Can't we omit the two above conditions (and therefore the $modules argument and always check module handler. I kinda feel that less logic in this method is desirable.

thunder.profile
```diff
@@ -239,12 +239,62 @@ function thunder_themes_installed($theme_list) {
...
+/**
+ * Check if provided triggering modules are one of just installed modules.
+ *
+ * This function is helper for thunder_modules_installed.
+ *
+ * @param array $modules
+ *   The list of the modules that were newly installed.
+ * @param array $triggering_modules
+ *   The list of triggering modules required for executing some action.
+ * @param bool $only_module_enable
```
alexpott: I still feel this is a confusing name. It seems to mean: return false during installation or config import if this flag is set.

ol0lll (dismissed):

alexpott (dismissed):

thunder.profile
alexpott: Might it make more sense to check this in the calling code because then it is in context of where it is happening?

thunder.profile
mtodor: We need `$modules` because we have to check if "hook_modules_installed" is triggered because of one of the modules we want to react on.

`if (empty($triggering_not_installed_modules)) {` is there only for performance reasons. I can kill it, but then we have to do `getModuleList`, `array_keys` and `array_diff`.

thunder.profile
mtodor: I'll pull it out to separate function.

dbosen (approved):

PR #671, opened 2019-03-19
Issue #3044344: Add thunder_amp blocks and remove profile weirdness

chrfritsch:


themes/thunder_amp/config/optional/block.block.thunder_amp_powered.yml
```diff
@@ -0,0 +1,19 @@
+langcode: en
+status: true
+dependencies:
+  module:
+    - thunder
+  theme:
+    - thunder_amp
+id: thunder_amp_powered
+theme: thunder_amp
+region: header
```
alexpott: Interesting region for a powered by block.

alexpott: Shouldn't this be in the footer?

alexpott: I guess we're maintaining the behaviour of block_theme_initialize() which places this in the header because thunder_amp doesn't have a footer_fifth region - but I wonder if we should take the opportunity to fix this.

alexpott (changes requested):
Should we add some test coverage to \Drupal\Tests\thunder\FunctionalJavascript\ModuleIntegrationTest::testAmpIntegration() that the blocks are appearing as expected as well?

ol0lll (changes requested):
When installing via gui and selecting amp feature in install form 12 default blocks are created, should be only the 4 exported ones.

chrfritsch:
> When installing via gui and selecting amp feature in install form 12 default blocks are created, should be only the 4 exported ones.

I think we can not fix this. We already learned that during the installation optional config will not be installed. That means block_theme_initialize() is creating all the blocks :( We could "fix" it by moving the blocks into config/install of the theme, but then when you want to enable thunder_amp later and for some reason, the block module is not enabled, that will not work.

So I would propose to leave it as it now is and remove AMP integration from the installer.

chrfritsch:
Moved to Thunder 3

PR #673, opened 2019-03-19
Cleanup the thunder breadcrumb builder

chrfritsch:


alexpott:
I think we should be careful here. Core makes no BC promises about the constructor so making this change means that whenever \Drupal\system\PathBasedBreadcrumbBuilder() makes a constructor change we have to change to. See https://www.previousnext.com.au/blog/safely-extending-drupal-8-plugin-classes-without-fear-of-constructor-changes - especially https://3v4l.org/KSYat for a safer way with little API impact.

PR #683, opened 2019-04-08
Issue #3046666: Deleting the media types removes thunder media view

chrfritsch:

### DIFF
--- a/modules/thunder_media/src/EventSubscriber/ConfigEventsSubscriber.php
+++ b/modules/thunder_media/src/EventSubscriber/ConfigEventsSubscriber.php
@@ -34,7 +34,7 @@ class ConfigEventsSubscriber implements EventSubscriberInterface {
    */
   public function configDelete(ConfigCrudEvent $event) {
     $config = $event->getConfig();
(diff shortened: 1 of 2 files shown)


modules/thunder_media/src/EventSubscriber/ConfigEventsSubscriber.php
```diff
@@ -0,0 +1,42 @@
...
+
+  /**
+   * React to a config object being deleted.
+   *
+   * @param \Drupal\Core\Config\ConfigCrudEvent $event
+   *   Config crud event.
+   */
+  public function configDelete(ConfigCrudEvent $event) {
+    $config = $event->getConfig();
+    if ($config->getName() == 'views.view.thunder_media' && ($media_view = View::load('media'))) {
```
dbosen: I admit, I do not care often enough about his, but we should always use === instead of ==

dbosen (approved):
Small nit, otherwise fine

dbosen (changes requested):
Just a nit.

dbosen (approved):

PR #685, opened 2019-04-10
Issue #3045492 by mtodor: Backend performance testing

mtodor:


mtodor:
We will move this functionality into a separate module.

mtodor:
I have moved only required libraries for Browser Performance Measurement to separate module.

tests/src/Nightwatch/Commands/paragraphsAddEmpty.js
```diff
@@ -0,0 +1,43 @@
...
+ * @param {string} type
+ *   The paragraphs type.
+ * @param {int} position
+ *   The position where paragraph should be added.
+ *
+ * @return {object}
+ *   The 'browser' object.
+ */
+
+exports.command = function paragraphsAddEmpty(fieldName, type, position) {
```
alexpott, line 18: This is only used in the paragraphsAdd command - is there a way to not make it part of the global commands but just something in scope there.

tests/src/Nightwatch/Commands/paragraphsAdd.js
```diff
@@ -0,0 +1,155 @@
...
+
+/* eslint-disable func-names */
+exports.command = function paragraphsAdd(fieldName, type, position, config) {
+  const browser = this;
+
+  const fieldNameId = fieldName.replace(/_/g, "-");
+  const newParagraphPosition = position * 2;
+
+  // Text Paragraph.
+  if (type === "text") {
```
alexpott, line 47: I think that rather than this method being a big switch on type that having addTextParagraph, addImageParagraphs, addInstagramParagraph is neater and more expressive.

alexpott: Also why paragraphsAdd() and not addParagraph()? I guess maybe to group commands. Maybe we could do something like paragraphs.addText()... and performance.startMark() - I dunno my JS-fu is not good enough to know off-hand.

tests/src/Nightwatch/Commands/performanceMarkEnd.js
```diff
@@ -0,0 +1,42 @@
...
+    // Set spanId to current active span, if there is any.
+    span = browser.apmSpans.pop();
+
+    if (!span) {
+      return;
+    }
+
+    browser.setCookie({
+      domain: browser.apmDomain,
+      expiry: 3533274000,
```
alexpott, line 32: This is optional - do we really need it? Also why Thursday, 18 December 2081 09:00:00

tests/src/Nightwatch/Commands/performanceMarkStart.js
```diff
@@ -0,0 +1,34 @@
...
+
+  browser.perform(() => {
+    const span = browser.apmTrans.startSpan(markName);
+    span.setTag("branch", process.env.THUNDER_BRANCH);
+
+    browser.apmSpans.push(span);
+
+    browser.setCookie({
+      domain: browser.apmDomain,
+      expiry: 3533274000,
```
alexpott, line 25: As above - and the rest below - is this necessary?

tests/src/Nightwatch/Commands/thunderFillCKEditor.js
```diff
@@ -0,0 +1,33 @@
...
+ *   The element selector.
+ * @param {string} value
+ *   The HTML value that will be filled to CKEditor.
+ *
+ * @return {object}
+ *   The 'browser' object.
+ */
+
+/* eslint-disable func-names */
+exports.command = function thunderFillCKEditor(selector, value) {
```
alexpott, line 17: Not sure about the thunder prefix here

tests/src/Nightwatch/Commands/thunderScrollIntoView.js
```diff
@@ -0,0 +1,49 @@
...
+ *
+ * @param {string} selector
+ *   The XPATH selector for element.
+ *
+ * @return {object}
+ *   The 'browser' object.
+ */
+
+/* eslint-disable func-names */
+exports.command = function thunderScrollIntoView(selector) {
```
alexpott, line 17: I think we can use https://nightwatchjs.org/api/moveToElement.html here instead - no?

tests/src/Nightwatch/Commands/thunderSelectTag.js
```diff
@@ -0,0 +1,27 @@
+/**
+ * @file
+ * Sets tag for element. It can be existing tag or new one.
```
alexpott, line 3: I think this could be more explicit about what we're dealing with. I think this is about taxonomy terms but select and tag are very very overloaded language.

alexpott (changes requested):

tests/src/Nightwatch/Commands/thunderScrollIntoView.js
mtodor: Problem with existing `move to element` functionalities is that they move an element into view of the browser window, but overlay elements are still there. So when the driver does click it clicks on overlay element and the test fails. That's why we are using scroll into view - what is actually doing scroll into the middle of the view.

tests/src/Nightwatch/Commands/paragraphsAddEmpty.js
mtodor: After split of `paragraphsAdd.js` into multiple commands, this command is used in several places.

tests/src/Nightwatch/Commands/paragraphsAdd.js
mtodor: I didn't find in documentation of Nightwatch if I can group commands in folders, but after checking code, I found it's possible. So, I have restructured all commands in sub-folders.

tests/src/Nightwatch/Commands/thunderFillCKEditor.js
mtodor: I'll remove it. I have prefixed it because I didn't want to have potential command collisions with core commands if they are introduced one day. But, if that happens, we will fix it.

tests/src/Nightwatch/Commands/thunderSelectTag.js
mtodor: It's widget specific, so I'll move it into widget related namespace.

tests/src/Nightwatch/Commands/performanceMarkEnd.js
mtodor: Yes, we do not need this option. I'll remove it.

And why `3533274000`, well that would be my 100th birthday. 😄

tests/src/Nightwatch/Commands/performanceMarkStart.js
mtodor: 👍

mtodor:
@alexpott I have addressed all your comments.

I have moved PR to the new repository. You can find it here: https://github.com/thunder/thunder-distribution/pull/19

First commit is code that you have reviewed. Commits after are related to your review and some Elastic APM deprecation warnings.

I'll close this PR since it's obsolete.

PR #694, opened 2019-05-07
Fix creating and editing articles with content_moderation uninstalled

alexpott:

### DIFF
--- a/tests/src/FunctionalJavascript/ArticleCreationTest.php
+++ b/tests/src/FunctionalJavascript/ArticleCreationTest.php
@@ -120,6 +120,27 @@ class ArticleCreationTest extends ThunderJavascriptTestBase {
   }
 
   /**
+   * Test Creation of Article without content moderation.
+   */
+  public function testCreateArticleWithNoModeration() {
+    // Delete all the articles so we can disable content moderation.
+    foreach (\Drupal::entityTypeManager()->getStorage('node')->loadByProperties(['type' => 'article']) as $node) {
+      $node->delete();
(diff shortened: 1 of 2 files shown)


tests/src/FunctionalJavascript/ArticleCreationTest.php
```diff
@@ -119,6 +119,27 @@ public function testCreateArticle() {
       ->elementsCount('xpath', '//div[contains(@class, "field--name-field-paragraphs")]/div[contains(@class, "field__item")][9]//span[contains(@data-pin-id, "478085316687452268")]', 2);
   }
 
+  /**
+   * Test Creation of Article without content moderation.
+   */
+  public function testCreateArticleWithNoModeration() {
+    // Delete all the articles so we can disable content moderation.
+    foreach (\Drupal::entityTypeManager()->getStorage('node')->loadByProperties(['type' => 'article']) as $node) {
```
chrfritsch, line 127: Interesting. Locally I had to remove all the moderation states in order to be able to disable the content_moderation module. The content could stay for me.

alexpott: Yeah that would work... I just did it one way that came to mind.

chrfritsch: Ok funny. The uninstall form recommends deleting the moderation states, but removing the content also works.
Anyway...

chrfritsch (approved):
This looks good to me

PR #709, opened 2019-06-18
Issue #3065653 by chr.fritsch, alexpott, daniel.bosen, volkerk: Fix notices and warnings in thunder_updater upgrade path

chrfritsch:

### DIFF
--- a/modules/thunder_updater/thunder_updater.install
+++ b/modules/thunder_updater/thunder_updater.install
@@ -91,7 +91,7 @@ function thunder_updater_update_8001() {
     $update_helper_checklist = $config_factory->getEditable('checklistapi.progress.update_helper_checklist');
 
     $items = $thunder_updater_checklist->get('progress.#items');
-    if (!empty($items) && is_array($items)) {
+    if (is_array($items)) {
       foreach ($items as $update_id => $update_info) {
         $update_helper_checklist->set('progress.#items.thunder:' . $update_id, $update_info);
       }
@@ -109,6 +109,13 @@ function thunder_updater_update_8001() {
 
   // Finally remove Thunder Updater!
   $module_installer->uninstall(['thunder_updater']);
+
+  // @todo This can be removed when
+  //   https://www.drupal.org/project/drupal/issues/3063734 lands.
+  // Remove thunder_updater from the update schema list.
+  register_shutdown_function(function () {
+    \Drupal::keyValue('system.schema')->delete('thunder_updater');
+  });
 
   return $update_logger->output();
 }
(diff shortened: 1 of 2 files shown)


modules/thunder_updater/thunder_updater.install
```diff
@@ -89,8 +89,12 @@ function thunder_updater_update_8001() {
     // Migrate Checklist API configuration.
     $thunder_updater_checklist = $config_factory->getEditable('checklistapi.progress.thunder_updater');
     $update_helper_checklist = $config_factory->getEditable('checklistapi.progress.update_helper_checklist');
-    foreach ($thunder_updater_checklist->get('progress.#items') as $update_id => $update_info) {
-      $update_helper_checklist->set('progress.#items.thunder:' . $update_id, $update_info);
+
+    $items = $thunder_updater_checklist->get('progress.#items');
+    if (!empty($items) && is_array($items)) {
```
dbosen: is_array() suffices as check

dbosen (dismissed):

chrfritsch:
Here is Travis using this PR in the upgrade testing https://travis-ci.com/thunder/thunder-distribution/builds/117658182.

alexpott (dismissed):
Looks good to me. Hopefully the core fix lands sometime.

chrfritsch:
This is not ready. We have to move the shutdown function to a new update hook because the old one already ran for everyone.

modules/thunder_updater/thunder_updater.install
```diff
@@ -106,5 +110,12 @@ function thunder_updater_update_8001() {
   // Finally remove Thunder Updater!
   $module_installer->uninstall(['thunder_updater']);
 
+  // @todo This can be removed when
+  //   https://www.drupal.org/project/drupal/issues/3063734 lands.
+  // Remove thunder_updater from the update schema list.
+  register_shutdown_function(function () {
```
alexpott: Needs to move to a new update function because this has occurred already. However - if the module is uninstalled we can't do the update here because the module has been uninstalled. So we need this here - but also maybe an update in the profile that does the clean up.

chrfritsch: I think we only need it in the profile, because of update dependencies we ensure that the thunder_updater hook runs at the very beginning

chrfritsch:
https://travis-ci.com/thunder/thunder-distribution/builds/117775696 shows we don't have a notice anymore. So I think this is done now

ol0lll (approved):
Tested in upgrade path to thunder 3: No more missing module warning.

PR #711, opened 2019-06-27
Issue #3059646: Consider upgrading field_group module

ol0lll:


ol0lll:
The field_group module provides an update hook, setting the 'region' key for each field_group.

ol0lll:
* Missing state handling: https://www.drupal.org/project/field_group/issues/3013395
* Field group in paragraphs: https://www.drupal.org/project/paragraphs/issues/2907094

dbosen (changes requested):
The inline form display of image media entity is broken now. Can be seen, when you open an image entity paragraph for editing, all the input fields are differently layouted than before.
Could also be the case on other form displays, but I did not find any.

ol0lll:
Patch from https://www.drupal.org/i/2907094 fixes field_groups which are directly in paragraphs, a use case we don't implement.
Patch from https://www.drupal.org/i/2986704 fixes field_groups in IEF - applied that.

composer.json
```diff
@@ -109,7 +113,7 @@
         "drupal/entity_browser": "1.7",
         "drupal/entity_reference_revisions": "^1.0",
         "drupal/fb_instant_articles": "^1.0",
-        "drupal/field_group": "^1.0",
+        "drupal/field_group": "^3.0",
```
chrfritsch: We have to fix the version.

drupal-org.make
```diff
@@ -48,7 +48,7 @@ projects[entity_reference_revisions][version] = 1
 projects[fb_instant_articles][type] = module
 projects[fb_instant_articles][version] = 1
 projects[field_group][type] = module
-projects[field_group][version] = 1
+projects[field_group][version] = 3
```
chrfritsch: We have to fix the version.

chrfritsch: Field group and paragraphs patch are missing

chrfritsch (changes requested):

chrfritsch (dismissed):
Looks good to me

chrfritsch:
Let's wait for the new paragraphs release now. Then we can get rid of the additional paragraphs patch

